PR #18, opened 2024-04-26
Making base executor more general so I can make a non dask executor to run on LUMI

DanHarJor:
I am having a problem that I cannot get dask to run on LUMI. I have decided to switch to a new tactic of using the scanning feature of GENE. SO I am writing a new executor and parser in order to make this happen.

When I went to make an executor I noticed there is dask implimented in the start_runs function in the base class. I didn't want my child class to inherit any dask stuff, so I moved start_runs into the dask executor. Then I can create my own executor with it's own start_runs that doesn't have any dask in it. I also put start_runs in the local_dask_executor and imported some needed functions. 

Since this might have a merge conflict and affect your code I wanted to try and merge it to main as soon as possible.

I hope this is okay.

### DIFF
--- a/src/executors/DaskExecutor.py
+++ b/src/executors/DaskExecutor.py
@@ -3,6 +3,8 @@
 from dask.distributed import Client
 from dask_jobqueue import SLURMCluster
 from .base import Executor
+from .base import run_simulation_task
+from dask.distributed import as_completed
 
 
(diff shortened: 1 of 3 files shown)


fusionby2030:
Closed due to #20  achieving the same thing

PR #20, opened 2024-05-14
Feature/active learning

fusionby2030:
On going development...


fusionby2030:
Adding
- [x] Collecting of data after "simulator" runs (simulator can also be pool based)
- [ ] Scaling of data before hand 
- [ ] model training 
- [ ] QLK example of pool sampling

fusionby2030:
Also handling all of the TODOs....

fusionby2030:
- [ ] Diagnostics for active learning steps (model state dict, acquired points, model metrics), and initial random sample
- [ ] Tests & Unit tests for Active Learning Structure (for collaboration)
- [ ] Add non-pool active learning 
- [ ] Add SLURM version (cpu & GPU versions)
- [ ] Random vs Kernel Method Comparison (for baseline problem -> QLK -> TGLF) 
- [ ] Multi-output regression? (Sum determinant or sum of determinants)

fusionby2030:
And eventually push to main...

fusionby2030:
- [ ] move bmdal to a subdirectory within active learning 
- [ ] update the fact that path to bmdal needs to be set in PYTHONPATH

fusionby2030:
I think the branch is stable (I have tested running MISHKA, HELENA, TGLF, the Active Static Pool (QLK dataset and example `train.csv` found in tests)), even though some of the tests don't run, it would be good to go through this together before pushing to main. @lorenzozanisi @awlishar 

@lorenzozanisi , I tested the select_batch on a few different kernels & `selection_methods`, but wouldn't hurt to try more.

awlishar (approved):
Looks good to me:)

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+            self.mapping_column_indices[col_idx] for col_idx in self.input_keys
+        ]
+        self.output_col_idxs = [
+            self.mapping_column_indices[col_idx] for col_idx in self.target_keys
+        ]
+
+        self.train, self.valid, self.test, self.pool = data_split(
+            self.data, **self.data_args
+        )
+        self.train = self.train.sample(0)
```
lorenzozanisi, line 44: why this line?

src/samplers/activelearner.py
```diff
@@ -0,0 +1,318 @@
...
+        for _ in range(self.init_num_samples):
+            params = [
+                torch.distributions.Uniform(lb, ub).sample().item()
+                for (lb, ub) in self.bounds
+            ]
+            param_dict = dict(zip(self.parameters, params))
+            batch_samples.append(param_dict)
+        return batch_samples
+
+    def _get_new_idxs_from_pool(self, model, train, pool) -> torch.Tensor:
```
lorenzozanisi, line 110: I think it might be important to specify in the docstring that the pool and train data are expected to be normalised.

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+"""
+
+from .base import Parser
+import torch
+import pandas as pd
+from common import data_split, CSVLoader, HDFLoader, PickleLoader, apply_scaler
+
+
+class STATICPOOLparser(Parser):
+    """This is a dummy class as for the moment the sampler handles everything"""
```
lorenzozanisi, line 22: Perhaps it would be useful to specify in the docstring that the usage intended is that the data should be stored as attributes and prior to scaling

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+            )
+        return data
+
+    def get_column_idx_mapping(self) -> dict:
+        return {
+            col: idx
+            for col, idx in zip(self.data.columns, range(len(self.data.columns)))
+        }
+
+    def get_train_valid(self) -> tuple[torch.Tensor, torch.Tensor]:
```
lorenzozanisi, line 71: is this even needed anymore?

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def get_train_valid_test_pool(
+        self,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        return self.train, self.valid, self.test, self.pool
+
+    def get_train_valid_datasplit_from_self(
+        self,
+    ) -> tuple[tuple[torch.Tensor, torch.Tensor], tuple[torch.Tensor, torch.Tensor]]:
+        """returns normalized (x_train, y_train), (x_valid, y_valid)"""
```
lorenzozanisi, line 95: is this true? I don't think that `self.train` and `self.valid` are ever normalised (but I see that line 49 is uncommented. Is this a docstring error?

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+
+        self.train, self.valid, self.test, self.pool = data_split(
+            self.data, **self.data_args
+        )
+        self.train = self.train.sample(0)
+        self.train = torch.tensor(self.train.values)
+        self.valid = torch.tensor(self.valid.values)
+        self.test = torch.tensor(self.test.values)
+        self.pool = torch.tensor(self.pool.values)
+        # self.train, self.valid, self.test, self.pool, self.scaler = apply_scaler(self.train, self.valid, self.test, self.pool, scaler=None, op='transform')
```
lorenzozanisi, line 49: probably useful to delete this, as the current logic in the Executor requires the data in the Parser attributes to be unscaled

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+        valid: torch.Tensor,
+        test: torch.Tensor,
+        pool: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+        train, valid, test, pool, scaler = apply_scaler(
+            train, valid, test, pool, scaler=None, op="transform"
+        )
+        return train, valid, test, pool
+
+    def get_train_valid_test_pool(
```
lorenzozanisi, line 87: maybe best to rename to something like `get_data_from_self` so it's clear it's the unscaled data that you are getting

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+            self.train[:, self.output_col_idxs].float(),
+        )
+        x_valid, y_valid = (
+            self.valid[:, self.input_col_idxs].float(),
+            self.valid[:, self.output_col_idxs].float(),
+        )
+
+        return (x_train, y_train), (x_valid, y_valid)
+
+    def make_train_valid_datasplit(
```
lorenzozanisi, line 107: specify in the docstring that the `train` and `valid` tensors are normalised already

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+            train[:, self.output_col_idxs].float(),
+        )
+        x_valid, y_valid = (
+            valid[:, self.input_col_idxs].float(),
+            valid[:, self.output_col_idxs].float(),
+        )
+
+        return (x_train, y_train), (x_valid, y_valid)
+
+    def update_pool_and_train(self, new_idxs):
```
lorenzozanisi, line 124: For future reference: a class that handles active learning where the pool is unlabelled needs to have a slightly different method here that takes in the queried inputs and the collected outputs. These must be unnormalised.

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+        return self.train, self.valid
+
+    def scale_train_val_test_pool(
+        self,
+        train: torch.Tensor,
+        valid: torch.Tensor,
+        test: torch.Tensor,
+        pool: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+        train, valid, test, pool, scaler = apply_scaler(
```
lorenzozanisi, line 82: `scaler` is presently unused

src/samplers/activelearner.py
```diff
@@ -0,0 +1,318 @@
...
+            y_train=y_train,
+            selection_method=self.selection_method,
+            sel_with_train=True,
+            base_kernel=self.base_kernel,
+            kernel_transforms=self.kernel_transform,
+        )
+
+        return new_idxs
+
+    def get_next_parameter(
```
lorenzozanisi, line 142: I think it might be important to specify in the docstring that the pool and train data are expected to be normalised.

src/executors/DaskExecutor.py
```diff
@@ -1,23 +1,238 @@
...
+                # print('Updated Pool size: ', len(self.sampler.parser.pool), 'Train size: ', len(self.sampler.parser.train))
+                # print(self.sampler.parser.return_dset_sizes())
+
+                # NOTE: Collect next data for training
+
+                train, valid, test, pool = (
+                    self.sampler.parser.get_train_valid_test_pool()
+                )
+                # rescale data and pool
+                train, valid, test, pool = (
```
lorenzozanisi, line 177: maybe better to call the variables `train_scaled`, `valid_scaled` and so on. Sorry for being so pedantic with this but I think this back and forth between scaled and unscaled variables could get confusing

src/samplers/activelearner.py
```diff
@@ -0,0 +1,318 @@
...
+            results.append(
+                {
+                    "input": self.parser.pool[idx, self.parser.input_col_idxs],
+                    "output": self.parser.pool[idx, self.parser.output_col_idxs],
+                    "pool_idxs": idx,
+                }
+            )
+        return results
+
+    def dump_results(self, base_run_dir: str) -> None:
```
lorenzozanisi, line 290: this should go in the ActiveLearnerBMDAL class

src/samplers/activelearner.py
```diff
@@ -0,0 +1,318 @@
...
+        self.write_metrics_to_file(metric_fname, self.metrics)
+        print(f"Saving final datasets to disk: {base_run_dir}")
+        for name, dset in zip(
+            ["train", "valid", "test"],
+            [self.parser.train, self.parser.valid, self.parser.test],
+        ):
+            fpath = os.path.join(base_run_dir, f"{name}.pt")
+            torch.save(dset, fpath)
+
+    def write_metrics_to_file(self, filename: str, metrics: list[float]):
```
lorenzozanisi, line 311: this should go in the ActiveLearnerBMDAL class

src/samplers/activelearner.py
```diff
@@ -0,0 +1,318 @@
...
+        Prints and saves final datasets to disk.
+
+        Args:
+            base_run_dir: Base directory path for saving.
+
+        """
+        print(100 * "=")
+        print(f"Final Results for {self.selection_method}")
+
+        metric_fname = os.path.join(base_run_dir, "final_metrics.txt")
```
lorenzozanisi, line 301: I think the full filename should be given, otherwise every time yo run with a different acquisition the file gets overwritten

lorenzozanisi: in fact, if using MLFlow or wandb or similar, better to pass the ID of the run and save that locally for ease, then use the API to figure out the metadata

src/parsers/STATICPOOLparser.py
fusionby2030: The sampler gathers initial points via `sampler.get_initial_parameters()`, so the training set should be empty before hand. Alternative would be to have `get_initial_parameters` do nothing for the static pool active learner.

src/parsers/STATICPOOLparser.py
fusionby2030: I don't think that is necessary as the function takes arbitrary train/valid tensors, so one could run `train, valid = self.get_train_valid() -> self.make_train_valid_dataset(train, valid)`. But we should consider documenting the executor workflow a bit more.

src/parsers/STATICPOOLparser.py
fusionby2030: I am not sure I understand, as the `update_pool_and_train` uses the unormalized (self) quantities.

src/parsers/STATICPOOLparser.py
fusionby2030: Could save it?

src/executors/DaskExecutor.py
fusionby2030: Was wanting to keep them the same name as to reduce memory usage...

src/samplers/activelearner.py
fusionby2030: I agree, but let's discuss what should be saved per active learning step and the data structure that should hold that.

src/parsers/STATICPOOLparser.py
lorenzozanisi: yeah I think the second option is cleaner

src/parsers/STATICPOOLparser.py
lorenzozanisi: yeah my point is more about the fact that `self.pool[:,col_idxs]` is not defined (or it's defined as NaN or whatever else is recognisable) when the pool doesn't have labels.

README.md
```diff
@@ -53,8 +53,7 @@ Although we will likely lint with `flake` so don't worry too much about it.
 
 The `tests` folder contains unit tests. These can be run manually by using the command:
```
lorenzozanisi, line 55: Somewhere in the README there should be a line that states that `bmdal` is a submodule and must be initialised with `git submodule update --init --recursive`

src/parsers/STATICPOOLparser.py
```diff
@@ -0,0 +1,149 @@
...
+            train, valid, test, pool, scaler=None, op="transform"
+        )
+        return train, valid, test, pool
+
+    def get_train_valid_test_pool(
+        self,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        return self.train, self.valid, self.test, self.pool
+
+    def get_train_valid_datasplit_from_self(
```
lorenzozanisi, line 92: is this actually ever used? I don't think the logic in the Executor requires this. Maybe we should remove this method

lorenzozanisi (approved):

PR #22, opened 2025-01-07
HELENA BETA ITERATOR

aejarvin:
HELENAparser.py and HELENArunner_beta.py modified/created to generate HELENA equilibria with a target beta_N through iteration.


awlishar:
@aejarvin @fusionby2030 How about moving the HELENArunner_beta functionality into the HELENArunner? So we have only one HELENArunner class but with the option of beta iteration?

aejarvin:
> @aejarvin @fusionby2030 How about moving the HELENArunner_beta functionality into the HELENArunner? So we have only one HELENArunner class but with the option of beta iteration? (I could do it and add some documentation)

Sounds good to me. I generated a different file originally to not interfere with the HELENArunner, while developing, but merging the two is indeed the clean way to go forward.

fusionby2030 (approved):

awlishar (approved):

PR #23, opened 2025-01-10
DREAM interface for enchanted-surrogates

aejarvin:
Standard DREAM running interface for enchanted-surrogates.


awlishar (approved):

fusionby2030 (approved):
Looks good. I am curious if DREAM will clog up the rundir or if it still outputs in the highest order rundir.

PR #27, opened 2025-02-05
Feature/output parser

DanHarJor:
A few fixes with the simple config file for n_workers. I added some files for running on lumi. I added the ability to include an output directory in the executor and have the output from the runners saved in a file there. At least in the sequential mode and not active learning mode.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - 'feature/**'
+      - 'Feature/**'
   pull_request:
     types: [opened, reopened]
 jobs:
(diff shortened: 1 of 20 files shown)


awlishar (approved):
Tested a simple local run and a batch HELENA run on mahti 👍

awlishar:
The SLURMcluster from dask_jobqueue launches a number of batch jobs defined with cluster.scale(x) (x: in the config files have been called either n_jobs or num_workers, but I think n_jobs is a better name). So the ".scale(...)" is necessary at least for my H+M workflow. 

The n_workers in the worker arguments then defines the number of workers within that batch job (I don't define that since I use only one worker per batch job). So that is different from n_jobs.

fusionby2030:
I am guessing we don't have any tests for trying it on `SLURM`?

fusionby2030 (approved):

PR #31, opened 2025-02-19
Feature/epedconstraint

awlishar:
- EPED constraint added as option in HELENA sampler and runner
- Beta iteration options added in HELENA sampler and runner
- Pedestal width scan options added in HELENA sampler and runner
- General improvements to HELENA and MISHKA samplers and runners


fusionby2030 (approved):
The lack of tests is a bit concerning but okay :D

PR #33, opened 2025-02-28
Test structure modification -> separate tests for local vs HPC

DanHarJor:
…hed from machine specific tests. I changed the pytest.yml to reflect this change. Also some of the test were not being trggered before so we might get more errors now. I believe it was active learning ones that were not being triggered.


awlishar (approved):
Looks good to me 👍

fusionby2030:
Nope not yet.

PR #34, opened 2025-03-03
DO NOT ACCEPT THIS PULL REQUEST IT IS JUST FOR A QUICK SANITY CHECK O…

DanHarJor:
…N THE PREVIOUS STATUS OF THE TESTS

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # enchated-surrogates
 
+sanity-check
+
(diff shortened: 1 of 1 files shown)


DanHarJor:
It was only to prove to me that the automated checks were working before my alterations. They were not working for me on LUMI for a fresh clone. I found out this was because the submodule bmdal was not loaded.

PR #36, opened 2025-03-08
Improve Error output when the simulator_run fails within a dask worker

DanHarJor:
Previously, when `run_simulation_task` was called within a DASK worker, there was a possibility that a run failed (for a given runner/parser combo) due to some error in the runner (it could be python or could be not happy handling of exits by a code),  and it was difficult to trace the error via STD out, as the worker would just fail, and the nanny would keep existing. 

now `run_simulation_task`  prints errors to std out(which can be found in the sbatch out file) instead of cryptically failing.

### DIFF
--- a/src/executors/base.py
+++ b/src/executors/base.py
@@ -21,7 +21,7 @@ import shutil
 from abc import ABC, abstractmethod
 import uuid
 import runners
-
+import traceback
 from dask.distributed import print
 
 # TODO: move to seperate file, tasks.py?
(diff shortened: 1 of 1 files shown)


fusionby2030 (approved):

fusionby2030:
I think one should handle their own errors in each specific runner, and we should avoid these kind of catch all errors in the simulation task but okay I guess this is okay. 

Still, when writing a runner you should have a good grip on what are the different failure modes, and put there also try-catches.

PR #37, opened 2025-03-08
Feature/helena lumi

DanHarJor:
I added some changes to the runner and parser of HELENA. 

Runner main change is that it can take a BETAN target as a parameter rather than being sampled

Parser so that it has the noKBM writer. 

I changed the beta itteration mechanisim, it is a linear prediction that corrects some mistakes of the previous one. It is also more clear that it is a linear prediction.


DanHarJor:
There is an issue with the beta iterator formula. This was causing the next beta to be very large and causing Helena to crash.

I re-implimented the linear prediction based beta iterator and here is an example print out of its performance:
BETA ITERATION 0.0: target=2.555, current=0.596
BETA ITERATION 0.1: target=2.555, current=0.72
BETA ITERATION 0.2: target=2.555, current=2.472
BETA ITERATION 0: target=2.555, current=2.5749999999999997
BETA ITERATION 1: target=2.555, current=2.557
BETA ITERATION 2: target=2.555, current=2.554
BETA ITERATION 3: target=2.555, current=2.555
BETA ITERATION FINISHED.
Target betaN: 2.555
Final betaN: 2.555
Number of beta iterations: 4

awlishar:
Change input_value_1/2 to more descriptive :) also add method documentation

awlishar:
New dependencies need to be added to requirements file if they are not there

awlishar:
@DanHarJor have you tested the changes in the executor on Mahti for example or for the other runners/samplers?

DanHarJor:
There seems to be some issues when using a sampler and a normal enchanted surrogates workflow. I will do some further debugging.

DanHarJor:
I added a test where I take 16 sobel sequence samples of the pedestal and do betaN itterations on all of them.

I am happy with test because it is all we need for DEEPlasma purposes. Do you want to try it on Mahti or other tests?

I am happy to merge with main now.

runLUMI.py
```diff
@@ -0,0 +1,25 @@
...
+#       This python file should be converted to a bash script
+for env_var in profile['env'].keys():
+    os.environ[env_var] = profile['env'][env_var]
+    
+# os.environ['PATH'] = os.environ['PATH'] + profile['env']['ENCHANTED_ENV_PATH']
+#------------------------------------------------------
+print('debug',profile['env']['ENCHANTED_SRC_PATH'])
+sys.path.append(profile['env']['ENCHANTED_SRC_PATH']) 
+
+import run
```
awlishar, line 23: Imports not in the middle of script

src/runners/HELENArunner.py
```diff
@@ -517,69 +564,151 @@ def run_helena_with_beta_at1_iteration(self, run_dir, beta_target):
...
         if self.beta_iterations_afp:
-            self.parser.modify_fast_ion_pressure("fort.10", 0.0)
+            self.parser.modify_fast_ion_pressure("fort.10", self.beta_iteration_starting_value_1)
         else:
-            self.parser.modify_at1("fort.10", 0.0)
+            self.parser.update_at1("fort.10", self.beta_iteration_starting_value_1)
+        
+        
         subprocess.call([self.executable_path])
+        import shutil
```
awlishar, line 594: Don't put imports in the middle of functions

tests/LUMI_tests/helena_old.py
awlishar, line 1: Is it necessary to have an "_old" test file? The old version should be retrievable from the git history

src/runners/HELENArunner.py
awlishar, line 1: Let's both (@DanHarJor and me) test this version before merging

awlishar (commented):
Needs testing and some code fixes

PR #39, opened 2025-03-19
Feature/multisimulation executor simulation pipeline

DanHarJor:
The old executor has been kept: DaskExecutor for legacy support

The executor functionality is now split:
DaskExecutorSimulation: This can be used when only a single runner is needed.
It only handels the creation of the cluster, getting samples from the sampler and sending runs to the cluster
It does not handle active learning or batch submission etc

DaskExecutorSimulationPipeline:
Within this multiple DaskExecutorSimulation objects will be defined.
Multiple PipelineParser functions will be defined

It handles the consecutive running of the runners defined for each DaskExecutorSimulation object and parsing the connection to the next runner with a PipelineParser function. 

TODO:
DaskExectuorActive:
This will handle running either a DaskExecutorSImulation or a DaskExecutorSImulationPipeline itteratlivly in an active learning loop.

Main Points for Adoption.
See simple_pipeline_config and simple_runner
# 1 
Your runners single_code_run() must take these inputs
single_code_run(self, run_dir: str, params=None)

if type(params)==type(None): [THEN RUN CODE AS IF RUN DIRECORY IS ALREADY SET UP TO GO]

# 2
When writing the configs file your runner must be declared within a DaskExecutorPipeline, as an argument. 
 
# 3
When writing functions to parse output from one code to input of another code you must define a python file for example MyPipelineParser.py and within the file functions NOT CLASSES, eg, def code1_to_code2():
Then these functions need to be imported in the __init__.py file of parsers. eg:
enchanted-surrogates/src/parsers/__init__.py
from .MyPipelineParser  import code1_to_code2

These functions must take in the arguments:
def code1_to_code2(last_out_dir, next_run_dir, run_id, future):

# 4
There may be other key points for implementation. Will work them out as we go. Have a look at the simple example or ask me if something isn't working.

### DIFF
--- a/src/runners/SIMPLErunner.py
+++ b/src/runners/SIMPLErunner.py
@@ -22,43 +22,44 @@ from parsers import SIMPLEparser
 class SIMPLErunner(Runner):
     """
     Runner class for simple test scenarios.
-
     Attributes:
         executable_path (str): The path to the executable binary.
-
     Methods:
         __init__(executable_path: str, *args, **kwargs)
             Initializes the SIMPLErunner object.
         single_code_run(params: dict, run_dir: str) -> str
             Runs a simple test program like a bash script.
-
     """
-
     def __init__(self, executable_path: str, *args, **kwargs):
         """
         Initializes the SIMPLErunner object.
-
         Args:
             executable_path (str): The path to the executable binary.
-
         """
         self.parser = SIMPLEparser()
         self.executable_path = executable_path
 
-    def single_code_run(self, params: dict, run_dir: str):
+    # in the pipeline executor we do not need to call this.
+    def parse_params(self, params:dict, run_dir:str):
+        self.parser.write_input_file(params, run_dir)
+    
+    def single_code_run(self, run_dir: str, params=None):
         """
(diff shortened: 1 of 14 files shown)


src/runners/SIMPLErunner.py
```diff
@@ -17,48 +17,55 @@
...
         self.parser = SIMPLEparser()
         self.executable_path = executable_path
+        self.type = kwargs.get('type')
 
-    def single_code_run(self, params: dict, run_dir: str):
+    # in the pipeline executor we do not need to call this.
+    def parse_params(self, params:dict, run_dir:str):
+        self.parser.write_input_file(params, run_dir)
+    
+    def single_code_run(self, run_dir: str, out_dir:str, params=None):
```
fusionby2030, line 49: This should stay being `run_dir, params` and not take `out_dir`. If you want to specify an `out_dir` then one can do it themselves in the parser and call it `single_code_run`.

fusionby2030 (changes requested):
Just the `single_run` other than that looks interesting!

fusionby2030 (approved):

PR #41, opened 2025-03-27
Bug fixes and extra return in run.main

DanHarJor:
Added hypercube and sobel to __init__.py

I removed out_dir from tasks.py

I added a return in run.main, so the sampler and executor are returned. This allows tests to check that they have the correct values.


DanHarJor:
I added a last commit, I forgot to take the out_dir out of the run_simulation_tasks in tasks.py.

So this is an important commit to merge with main also.

PR #43, opened 2025-03-31
Feature/gene simulation lumi

DanHarJor:
I added gene runner and parser. It does not run a GENE scan but a single local simulation in one ky. I made a ky sampler that will sample ion and electron kys seperatly.

### DIFF
--- a/deeplasma_extra_requirements.txt
+++ b/deeplasma_extra_requirements.txt
@@ -1,0 +1,2 @@
+#TPED
+xarray
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -21,10 +21,10 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
+        git config submodule.src/parsers/submodules/IFS_scripts all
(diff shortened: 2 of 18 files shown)


DanHarJor:
By adding update = non in .gitmodules I got the pytest not to load these submodules.

The next issue is that when any enchanted surrogates imports happens and the __init__.py files get ran the submodules are then not there for the parser and sampler imports.

There is an issue related to this which removes everything from the __init__.py and imports each thing individually as we need them. This would solve my pytest issue here aswell.

deeplasma_extra_requirements.txt
awlishar, line 1: @DanHarJor Why can't this be in the general requirements.txt file?

awlishar (commented):
Looks OK, can you perhaps test that this version still works, since it has been a while since the PL was created?

deeplasma_extra_requirements.txt
DanHarJor: The idea was that not every project needs all the packages. So if a user can't install the packages because of dependancy issues or whatever then they don't have to. I talked to Adam about this and he has another strategie for dealing with this. 

Lets not merge this one. The others are okay to merge though

PR #45, opened 2025-04-07
Feature/targeted imports

### LINKED ISSUE #42: Imports taking too long / Importing everthing when only small scope is used
Essentially we need to use the information in the config file to only import what is necessary.

Steps to accomplish this:
# 1: 
Remove all imports from \_\_init\_\_.py for parsers, samplers, executors and runners
When these are here it is impossible to import anything without importing everything.
THIS IS A LITTLE DISRUPTIVE AS SOME IMPORTS WOULD NEED TO CHANGE, FOR EXAMPLE WHEN IMPORTING THE PARSERS IN EACH RUNNER. 

# 2: 
We can use strings to import what we need with the code:
for example for executors:
`executor = getattr(importlib.import_module(f'executors.{executor_type}'),executor_type)(**executor_args)
`

# 3:
The previous command required that the string executor_type is both the module name and the class name. This is currently true for all the executors,  parsers and runners but NOT for the samplers.

## 3.1 
We alter the samplers so there is only one file per class and the file name matches the class name. activelearner.py and rand.py would need to be split into two files since they contain two sampler classes each. BETTER FOR USER EXPERIENCE, MIGHT BE MORE DISRUPTIVE, ONLY ONE CLASS PER FILE

# OR

## 3.2 
When specifying the sampler in the config file we specify the module and class_name in different attributes. This allows us to use the command in 2 when the module and class_name are different. MAKES THE USER EXPERIENCE A BIT WORSE, MIGHT BE LESS DISRUPTIVE. ALLOWS FOR MANY CLASSES TO BE IN ONE FILE

# Which do you think is best?

I did a quick test and it works very well, the imports take no time at all when we only import what we need. Maybe some configurations will take longer than others but they will all surely be faster than if they all imported everything.

DanHarJor:
I added the code for the target imports.

This also contains the bug fixes from the other pull request:
https://github.com/DIGIfusion/enchanted-surrogates/pull/41


DanHarJor:
This will break code when

1. you import a parser, sampler, runner, executor without specifying the module/file name because it used to be specified in the __init__.py

eg 
now this does not work
from parsers import HELENAparser
now we need to do thit
from parsers.HELENAparser import HELENAparser

2. If you define a sampler, runner or executor object based on the config file with the old method:
args.executor['runner_args'] = getattr(args, 'runner', None)
instead of the new method that handeles the import aswell:
executor = getattr(importlib.import_module(f'executors.{executor_type}'),executor_type)(**args.executor)

src/executors/tasks.py
fusionby2030, line 1: Until we try this on different HPC systems, I think we should create the runner in the `run_simulation_task`, because if I recall, there was an issue with pickling more complex objects than standard python stuffs, but I could be wrong here. Perhaps some MAHTI or other HPC tests to verify this?

fusionby2030 (commented):
except for runnner things looks like a good change to improve importing stuffs.

DanHarJor:
@fusionby2030 
I made the requested revert

fusionby2030 (approved):

src/executors/tasks.py
DanHarJor: After thinking, it is probably better we make a new runner object in each worker anyway. Then each workers runner can have their own attributes which are different depending on how the run is going. The babysitting of the codes. 

Thanks for catching this.

PR #46, opened 2025-04-11
bug/run_dir

DanHarJor:
I found a bug in DaskExecutor that wasn't detected by the tests so I added a test capable of finding it then fixed the bug. It was only making one run_dir because I accdidently put the run_dir creation out of the loop


awlishar (approved):
Looks OK 👍

PR #49, opened 2025-06-04
Dreamliner mahti

aejarvin:
Here is a pull request with the latest BO sampler implementation for DREAM.


aejarvin:
I'll clean up the DaskExecutor.py a bit first before PR.

PR #53, opened 2025-09-02
Dev/feature/dask executor

DanHarJor:


awlishar:
Fixed mainly the imports and dependencies. Both tests ran successfully on mahti for me. 
Note that the test_dask_executor.py still needs user input to run and will always fail in GA. However, I think that is fine for now. I will still checkout the merge conflicts.

PR #57, opened 2025-09-08
Dev/feature/dask executor precise imports

DanHarJor:


DanHarJor:
Hello, I mostly wanted to add the precise imports and fix the monkey patch in dask_executor. I had to do some alterations to the rest of the code to keep things consistent, allow the sys.path.append method instead of pip install . -e (as this won't work on HPC read only python environments). Then I hit some bugs with the test working on LUMI but not on git workflows. Now it is all Peachy but the commit tree might be a bit messy. Feel free to question me about any changes.

awlishar:
What do you mean with the "read only python environments"? All dependency packages need to be installed anyway

DanHarJor:
> What do you mean with the "read only python environments"? All dependency packages need to be installed anyway

In HPC because python enviorments have many files and overload the filesystem it is best practise to put your python enviroment into a squashfs file. Then when you use squashfs software to mount the enviroment directory it only grabs one large file from the lustre filesystem once and then when python uses the enviroment squashfs software handles getting each individual file and the lustre filesystem does nothing.

LUMI has made this a more user friendly and robust with containers:
https://docs.lumi-supercomputer.eu/software/installing/container-wrapper/

The drawback is that then the python enviroment is read only. So if you do pip install -e . inside es

Then when you go to edit the code your edits wont be available in the python environment because it is read only.
So you need to do sys.path.append(path/to/src) to be HPC friendly

I added some functions to make this a bit smoother so different users don't need to input their own path/to/src.

DanHarJor (commented):
reviewed by Daniel and Ammanda

awlishar (approved):
Daniel walked me through the changes, we will push to develop and check that the plugin structure still works.

PR #65, opened 2025-09-10
small updates

DanHarJor:


DanHarJor:
some updates to 
add a gneric test to see if any plugins are installed
Force users to have a return a dict from single code runner that at least has 'success' as a key

PR #97, opened 2025-10-18
feature/hdf5_data_pack

DanHarJor:


DanHarJor:
This is a temporary fix to the many files problem.

It only creates the hdf5 files at the end. Which means that the many files need to exist at some point. The hdf5 files should be made gradually.

awlishar:
can you change the file name from hdf5.py to something like hdf5_helper.py since there also exists an hdf5 package

DanHarJor:
I added the executor updates as some are relevant to the hdf5 packing. I thought they were already in develop.

DanHarJor:
I had to add the run_simulation_task updates for the executor to work. This includes more robust error handling.

DanHarJor:
I added the necessary changes to features/core branch

PR #106, opened 2025-10-22
Lower Python version requirement to >=3.8

DanHarJor:


awlishar:
@DanHarJor If you change the required python version, you need to test that. I added 3.8 to the tests and as you can see above, both tests fail.

DanHarJor:
Nice catch, thanks. I had to lower it for SDCC machine, everthing I need to work still seems to work for me. But we don't need to change it for everyone.

PR #110, opened 2025-10-31
Feature/csv sampler.py

DanHarJor:
This is for Otso and his ASCOT marker workflow.


DanHarJor:
It was too messy here with the other files added somehow. I have now put it in another pull request, feature/samplers

Issue #109, opened 2025-10-24
Resource usage stats


DanHarJor:
## User story

As a user, I choose to use the Dask Executor with SLURM clusters. When preparing my run, I chose how many "workers"/batch jobs I need and how many cores, memory etc that I allocate for each. After my run has finished, I want to get an overview of how much of those resources that my workers actually used. I know I can check it with the following command:

```
seff <jobid>
```

and get an answer looking something like this:

```
Job ID: 123456
State: COMPLETED (exit code 0)
Cores: 4
CPU Utilized: 01:01:32
CPU Efficiency: 98.6%
Memory Utilized: 1.02 GB
Memory Efficiency: 85.3%
```

but this is not efficient for me to do if I have many batches in my run.

## Description

One of the log files should contain a breakdown of how many cpuh and memory were used. This should include the total used, what was used by each dask cluster that was ran.

## Acceptance Criteria

- [ ] There is a log file (`logger.info`) that summarises the resource usage of the enchanted surrogates run
- [X] It includes the total cpuh used by each Dask cluster.
- [X] It includes the total cpuh used by non dask cluster resources, specifically,the main python thread ("nanny") that is ran in an sbatch.

## Optional

- [ ] Implement a similar resource logging for the other executors

PR #117, opened 2025-11-17
Resource usage stats

JaaskelainenL:
This regards the issue #109 

Currently this uses printing instead of logging because another task is focusing on the logging part

### DIFF
--- a/src/enchanted_surrogates/executors/dask_executor.py
+++ b/src/enchanted_surrogates/executors/dask_executor.py
@@ -80,11 +80,11 @@ class DaskExecutor(Executor):
     def get_slurm_usage_info(self, job_id=None):
         """
         Params:
-            job_id (int): If you wish to only find the slurm usage info from one specific job pass this
+            job_id (list[int]): If you wish to only find the slurm usage info from one specific job pass this
         Returns:
             list: dictionary containing the output info from running seff
         """
-        job_ids = [job_id] if job_id else self.get_all_dask_job_ids()
+        job_ids = job_id if job_id else self.get_all_dask_job_ids()
         job_info = []
 
         for job_id in job_ids:
@@ -92,6 +92,9 @@ class DaskExecutor(Executor):
                 result = subprocess.run(['seff', job_id], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
                 output = result.stdout
                 seff_lines = output.splitlines()
+
+                if len(output.strip()) == 0:
+                    continue
                 
                 cpu_time = self.find_line_in_seff_output(seff_lines, "CPU Utilized:")
                 cpu_efficiency = self.find_line_in_seff_output(seff_lines, "CPU Efficiency:")
@@ -124,7 +127,7 @@ class DaskExecutor(Executor):
         try:
             jobs = []
             result = subprocess.run(['squeue', '--me'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
-            dask_lines = [line for line in result.stdout.splitlines() if 'dask-wor' in line]
+            dask_lines = [line for line in result.stdout.splitlines() if 'sys/dash' not in line and "SURROGAT" not in line]
 
             if not dask_lines:
                 print("No Dask jobs found in queue.")
@@ -330,6 +333,7 @@ class DaskExecutor(Executor):
         if not self.client:
             self.start_cluster()
         print('CLUSTER STARTED')
+        all_job_ids = self.get_all_dask_job_ids()
         all_futures = []
 
         while self.sampler.has_budget:
@@ -371,7 +375,7 @@ class DaskExecutor(Executor):
         with open(os.path.join(self.base_run_dir,'ENCHANTED.FINISHED'), 'w') as file:
             file.write(f'ENCHANTED.FINISHED, {__class__}')
 
-        job_info = self.get_slurm_usage_info()
+        job_info = self.get_slurm_usage_info(all_job_ids)
         total_cpu_time = sum(job['cpu_time_seconds'] for job in job_info)/3600
         print(f"Total CPU hours used: {total_cpu_time}")
 
--- a/src/enchanted_surrogates/executors/dask_executor.py
+++ b/src/enchanted_surrogates/executors/dask_executor.py
@@ -64,6 +64,82 @@ class DaskExecutor(Executor):
         self.client = None
         self.expected_number_of_workers = None
 
+    def get_slurm_usage_info(self, job_id=None):
+        """
+        Params:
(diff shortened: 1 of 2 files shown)


src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -64,6 +64,81 @@ def __init__(self, *args, **kwargs):
...
+                cpu_secs = hours * 3600 + minutes * 60 + seconds
+                
+                job_info.append({
+                    'cpu_time': cpu_time,
+                    'cpu_time_seconds': cpu_secs,
+                    'cpu_efficiency': cpu_efficiency,
+                    'memory_efficiency': memory_efficiency,
+                    'memory_used': memory_used,
+                    'job_id': job_id
+                })
```
laurisaksi, line 114: This could be extracted and you should write unit tests for this parsing.

laurisaksi (changes requested):
I left one comment about the unittest but otherwise looking good👍

laurisaksi (approved):
LGTM. We can improve the testing at some point.

Before merging wait for the second review.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -64,6 +64,85 @@ def __init__(self, *args, **kwargs):
...
+        """
+        Runs squeue to figure out all jobs from the cluster
+        Returns:
+            list: A list of Dask job IDs.
+        """
+        try:
+            jobs = []
+            result = subprocess.run(['squeue', '--me'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+            dask_lines = [line for line in result.stdout.splitlines() if 'sys/dash' not in line and "SURROGAT" not in line]
+
```
DanHarJor, line 131: Every dask worker has the name set by: #SBATCH -J dask-worker

This could be used to get all the dask workers.

The users make their own sbatch to launch run.py and they might not put 'SURROGAT' in the name.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -294,6 +374,18 @@ def start_runs(self):
...
             file.write(f'ENCHANTED.FINISHED, {__class__}')
+
+        job_info = self.get_slurm_usage_info(all_job_ids)
+        total_cpu_time = sum(job['cpu_time_seconds'] for job in job_info)/3600
+        print(f"Total CPU hours used: {total_cpu_time}")
+
+        cpu_ps = subprocess.run(["ps", "--no-headers", "-o", "etimes=", "-p", str(os.getpid())], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+        if cpu_ps.returncode == 0:
+            nanny_secs = int(cpu_ps.stdout.strip())
+            print(f"Total CPU time used by nanny: {nanny_secs/3600}")
```
DanHarJor: In dask the nanny is a process that runs on each worker node. There is some confusion about what the nanny is amongst my team so I suspect there has been in your discussions too.

We should call this the HEAD node of the cluster. In dask it is usually known as the scheduler node and it runs the:
scheduler process
main python process which also has the client.

The nanny process on each worker will communicate with the scheduler process regularly. 

So we should call it the HEAD node as this is general for any cluster that enchanted surrogates uses. 

So change this line to toal cpu time used by HEAD node:

DanHarJor (changes requested):
This is overall a nice implimentation for the dask executor.

As we are considering having more executors it would be good to have this functionality outside the dask executor.

We can potentially enforce that all jobs made by enchanted surrogates start or end with 'enc' so that they can be identified. 

for dask executor this would require adding '-J enc_dask_worker' to job_extra_directives before the cluster is made and test that it actually changes the job name.

It can then be called in run.py

Maybe making a more general one can be apart of the next sprint.

Just change the naming convention from nanny to head-node, and try not to rely on 'SURROGAT' being in the head node job name and I will approve this pull request.

src/enchanted_surrogates/executors/dask_executor.py
JaaskelainenL: is adding '-J enc_dask_worker' ok or should I figure out what the user has provided and add "enc_" as a prefix if it doesn't already have one?

DanHarJor: adding -J enc_dask_worker is fine. I don't expect the users to care much about the job name for the workers.

JaaskelainenL: done :+1:

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -127,7 +127,7 @@ def get_all_dask_job_ids(self):
         try:
             jobs = []
             result = subprocess.run(['squeue', '--me'], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
-            dask_lines = [line for line in result.stdout.splitlines() if 'sys/dash' not in line and "SURROGAT" not in line]
+            dask_lines = [line for line in result.stdout.splitlines() if 'sys/dash' not in line and "enc_dask_worker" not in line]
```
DanHarJor: This should be 

dask_lines = [line for line in result.stdout.splitlines() if 'sys/dash' not in line and "enc_dask_worker" in line]

'enc_dask_worker' should be in the line for it to return the lines of the workers.

DanHarJor (changes requested):
I think the is a small mistake, the current versions shouldn't find any dask worker jobs.

Can you check the jobs are actually names enc_dask_worker with 

squeue --me

Can you put the print of the usage stats in this pull request to show it works.

Best

Daniel

src/enchanted_surrogates/executors/dask_executor.py
JaaskelainenL: fixed

JaaskelainenL:
$ squeue --me
```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
           5521155 interacti SURROGAT  jaaskel  R       0:38      1 c3101
           5521149 interacti sys/dash  jaaskel  R       5:35      1 c3101
           5521160     small enc_dask  jaaskel PD       0:00      1 (Priority)
           5521159     small enc_dask  jaaskel PD       0:00      1 (Priority)
           5521158     small enc_dask  jaaskel PD       0:00      1 (Priority)
           5521157     small enc_dask  jaaskel PD       0:00      1 (Priority)
           5521156     small enc_dask  jaaskel PD       0:00      1 (Priority)
```

DanHarJor (approved):
The code looks good. If you have checked it works then you can merge. If you have time put a print of the usage stats in here as a comment to record it working.

Daniel

PR #118, opened 2025-11-17
Feature/enc 59 develop a release workflow

tsseppa:
Created pull request for documenting developing phases

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,6 +1,7 @@
 name: Pylint
 
 on: 
+  workflow_call:
   push:
     branches:
       - main
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,6 +1,7 @@
 name: Unit testing with Pytest
 
 on: 
+  workflow_call:
   push:
     branches:
       - main
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -8,9 +8,9 @@ on:
 jobs:
 
   lint:
-    uses: ./lint.yaml
+    uses: .github/workflows/lint.yaml
   test:
-    uses: ./test.yaml
+    uses: .github/workflows/test.yaml
   build:
     runs-on: ubuntu-latest
     needs: [lint, test]
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,1 @@
+# Generate release workflow file here
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,15 +14,18 @@ jobs:
   build:
     runs-on: ubuntu-latest
     needs: [lint, test]
+    strategy:
+      matrix:
+        python-version: ["3.10"]
     steps:
         - name: Checkout
           uses: actions/checkout@v6
           with:
             fetch-depth: 0
-        - name: Set up Python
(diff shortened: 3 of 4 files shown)


tsseppa:
# Documentation for creating release

Releases are accessed in [GitHub](https://github.com/DIGIfusion/enchanted-surrogates). Only users with write or push access to project can create release. Releases are created from `main` branch.

## Semantic versioning

Workflow uses GitHub actions to create Releases on GitHub. It runs automatically when you push a Git tag that matches the pattern `vX.Y.Z` where `X.Y.Z` is using [Semantic Versioning](https://semver.org/). It matches `major.minor.patch` type of semantic versioning:

1. Major `X` is increased when something backwards incompatible is created
2. Minor `Y` is increased when functionality is added in a backward compatible manner 
3. Patch `Z` is incremented when backward compatible bug fixes are made.

Increasing `X` resets `Y` and `Z`  to zero, and increasing `Y` resets `Z` to zero. Each increment is always one.

**Version should be incremented for each release**

## Triggering release

Currently workflow is triggered in two steps: first `git tag` and `git push` commands in development environment, then the release is created from [Releases subpage](https://github.com/DIGIfusion/enchanted-surrogates/releases/)






### Step 1: Local triggering

This requires user to create tag name `vX.Y.Z` according to semantic versioning described above. Workflow activates `.github/workflows/release.yaml`.


In `release.yaml` file linting and testing is run, then if those succeed package is built, and if build succeeds content of  `dist/*` folder is included in release. 

Local triggering is done in the branch, that is to be released.
Use below commands to trigger first part of release workflow:
```bash
git tag vX.Y.Z
git push origin vX.Y.Z
```

### Step: 2 GitHub release
Check on [Actions](https://github.com/DIGIfusion/enchanted-surrogates/actions) page, that tests, build and artifact creation succeeded.

Go to [Releases](https://github.com/DIGIfusion/enchanted-surrogates/releases) page to create a new release:

1. Click Draft a new release. 
2. Select the tag `vX.Y.Z` which was used for local triggering
3. Click Generate release notes and use it as a base
4. Modify release notes if necessary and publish

tsseppa:
# Documentation for testing the workflow

I am not sure if creating release should be internal information or something to be included in the website. I wrote above documentation in md that can be used in the website if wished. I will probably add above to `/docs/releases.md`

## Reasoning for architecture

Currently the developer secret `GITHUB_TOKEN` has not permissions required for creating a release , so this process cannot be automized fully. There could be other ways as well, which can be investigated if needed.

## How to test?

Delete first the current release from [Releases](https://github.com/DIGIfusion/enchanted-surrogates/releases) page. Then, enter to the git branch `feature/enc-59-develop-a-release-workflow`

Check if there are local tags with 
```bash
git tag
```
Delete local tags with 
```bash
git tag -d [tag name]
```
Delete remote tag (the one used for release just deleted)
```bash
git push origin --delete [tag name]
```

**Then, use the instructions given above to create the release**


## What could be implemented in the future?

1. More automatic release with more permissions given in `GITHUB_TOKEN`
2. Add PyPI intergation so instead of downloading .zip or cloning the repo, user can just run 
```bash
pip install enchanted-surrogates
```

dafughes (approved):
Nice work and good documentation!

Husseinhamsa (commented):
Näyttää oikealta

awlishar (approved):
Tested and works 👍

PR #126, opened 2025-11-24
Feature/core

DanHarJor:
From the branch I have been developing I took all the files used for core functionality of ES and put them here for review.


src/enchanted_surrogates/__init__.py
DanHarJor, line 1: I had issues with some plugins not loading and there was no indication of which one and why, so I added a print so I could see which one is loading, then if it fails silently while one is loading I know which one it is. There is also a try catch to show the name of which one that failed.

pyproject.toml
DanHarJor, line 1: Some of my new runners are fast and have plotting functions for users to see the real function. So I added matplotlib and plotly.

aejarvin:
This pull request must be split into smaller increments. There is no way we can review and merge a PR with 16 files changed and expect the transition to be smooth.

DanHarJor:
> This pull request must be split into smaller increments. There is no way we can review and merge a PR with 16 files changed and expect the transition to be smooth.

Yes, that is more or less what Amanda said. So I did split it into smaller pull requests, this one was left as a draft pull request. But closing it is fine too.

PR #136, opened 2025-11-24
Feature/runners

DanHarJor:
created from devfeature/new_samplers


aejarvin:
Split to smaller increments.

DanHarJor:
Same here, once the core has been implemented then this one will become a small pull request that just contains some new runners. It will only be a few files.

PR #143, opened 2025-11-25
feature/error_handeling

DanHarJor:
When an error happens in single_code_run, run_simulation_task catches it and returnes a different dictionary than if it had succeded. This then messes up the summary file, enchanted_dataset.csv. So I changed the executors to to make two summary files, one for succeded runs and one for failed, so they can have different headders. In run_simulation_task more error info is recorded that wasn't previously and this is optionally returned if the executor requests. Then the executor can make a file with the detailed error information for each failed run, currently this is a json file. There are other small changes, see comments in files changed

### DIFF
--- a/src/enchanted_surrogates/executors/dask_executor.py
+++ b/src/enchanted_surrogates/executors/dask_executor.py
@@ -331,7 +331,7 @@ TO AVOID THIS PLEASE ISSUE INCLUDE ANY TIMEOUTS IN YOUR RUNNER AND HANDLE EARLY 
                     shutil.rmtree(batch_dir)
                     break
                 
-                _futures = self.submit_batch(samples, base_run_dir=batch_dir, request_errors=True, rm_run_dir=~self.save_run_dirs)
+                _futures = self.submit_batch(samples, base_run_dir=batch_dir, request_errors=True)
                 futures = set(_futures)
                 del _futures
                 num_samp_in_batch = len(futures)
@@ -430,7 +430,7 @@ TO AVOID THIS PLEASE ISSUE INCLUDE ANY TIMEOUTS IN YOUR RUNNER AND HANDLE EARLY 
                 sample_run_dir = make_run_dir(base_run_dir=base_run_dir, prepend=self.runner_config['type']) 
             run_dirs.append(sample_run_dir)
             new_future = client.submit(
-                run_simulation_task, self.runner_config, sample_run_dir, sample_params, return_errors=request_errors, retries=0
+                run_simulation_task, self.runner_config, sample_run_dir, sample_params, return_errors=request_errors, rm_run_dir=~self.save_run_dirs, retries=0
             )
             futures.append(new_future)
             fut_to_rundir[new_future.key] = sample_run_dir
--- a/src/enchanted_surrogates/executors/dask_executor.py
+++ b/src/enchanted_surrogates/executors/dask_executor.py
@@ -1,16 +1,25 @@
 import os
+#!/usr/bin/env python3
+import glob
+import os
+            
 import subprocess
 import time
 import warnings
 import pandas as pd
+import json
 
 from dask_jobqueue import SLURMCluster
 from dask.distributed import LocalCluster
 from dask.distributed import Client, as_completed, wait, LocalCluster, get_worker, get_client
 from dask.distributed import print as dask_print
+from enchanted_surrogates.utils.time_format import format_sec
+
+from enchanted_surrogates.utils.precise_imports import import_runner
 
 from .base_executor import Executor
 from enchanted_surrogates.executors import simulation_task
+import shutil
 from enchanted_surrogates.utils.make_run_dir import make_run_dir
 from enchanted_surrogates.utils.precise_imports import import_sampler
 
@@ -51,18 +60,33 @@ class DaskExecutor(Executor):
         super().__init__(*args, **kwargs)
         print('INITIALISING DASK EXECUTOR')
         self.type = kwargs.get('type')
+        self.base_run_dir = kwargs.get('base_run_dir')
+        self.sampler_config = kwargs.get('sampler_config')
         if self.sampler_config:
+            self.sampler_config['base_run_dir'] = self.base_run_dir
             self.sampler_type = self.sampler_config.pop("type")
-            self.sampler = import_sampler(
-                type=self.sampler_type, sampler_config=self.sampler_config)
-        self.scale_n_jobs = kwargs.get('scale_n_jobs', 1)
+            self.sampler = import_sampler(type=self.sampler_type, sampler_config=self.sampler_config)
+        self.scale_n_jobs = kwargs.get('scale_n_jobs', None)
+        self.scale_n_jobs_min = kwargs.get('scale_n_jobs_min', None)
+        self.scale_n_jobs_max = kwargs.get('scale_n_jobs_max', None)
+        assert not (self.scale_n_jobs and (self.scale_n_jobs_min or self.scale_n_jobs_max)), 'EITHER scale_n_jobs OR scale_n_jobs_min/scale_n_jobs_max CAN BE SET, NOT BOTH'
+        
+        self.runner_config = kwargs.get('runner_config')
+        
         self.timeout = kwargs.get('timeout', 1e10)
+        self.run_error_log_path = os.path.join(self.base_run_dir, 'runs_error_log.jsonl')
         self.SLURMcluster_config = kwargs.get('SLURMcluster_config')
         self.LocalCluster_config = kwargs.get('LocalCluster_config')
         self.block_until_cluster_started = kwargs.get('block_until_cluster_started', False)  # for debugging purposes only
         self.cluster = None
         self.client = None
         self.expected_number_of_workers = None
+        self.current_batch = 0
+        self.save_run_dirs = kwargs.get('save_run_dirs', True)
+        self.submit_command = kwargs.get('submit_command', None)
+        
+        self.psudo_runner = import_runner(self.runner_config['type'], runner_config=self.runner_config)
+
 
     def start_cluster(self, slurm_out_dir=None):
         """
@@ -79,14 +103,9 @@ class DaskExecutor(Executor):
             Warning: If fewer workers than expected are started.
         """
         print('MAKING CLUSTER')
-        worker_logs_dir = None
-
         if self.SLURMcluster_config:
-            self.expected_number_of_workers = self.scale_n_jobs * int(self.SLURMcluster_config.get('processes',1))
-
             if not slurm_out_dir:
                 slurm_out_dir = os.path.join(self.base_run_dir,'worker_out_DaskExecutor')
-            worker_logs_dir = slurm_out_dir
             if not os.path.exists(slurm_out_dir):
                 os.makedirs(slurm_out_dir)
             jed = self.SLURMcluster_config.get('job_extra_directives')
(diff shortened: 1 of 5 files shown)


src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -219,11 +257,6 @@ def count_alive_workers(self):
                 count += 1
         return num_alive_workers
```
DanHarJor: scale functionality can be added later

DanHarJor (commented):
test

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -1,16 +1,25 @@
 import os
+#!/usr/bin/env python3
+import glob
+import os
+            
 import subprocess
 import time
 import warnings
 import pandas as pd
+import json
```
DanHarJor, line 6: Used for making an error summary file

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -1,16 +1,25 @@
...
 
 from dask_jobqueue import SLURMCluster
 from dask.distributed import LocalCluster
 from dask.distributed import Client, as_completed, wait, LocalCluster, get_worker, get_client
 from dask.distributed import print as dask_print
+from enchanted_surrogates.utils.time_format import format_sec
+
+from enchanted_surrogates.utils.precise_imports import import_runner
```
DanHarJor, line 14: Used to make a psudo_runner object that can be used to run a light post and pre processing function if the runners have one.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -1,16 +1,25 @@
...
 import time
 import warnings
 import pandas as pd
+import json
 
 from dask_jobqueue import SLURMCluster
 from dask.distributed import LocalCluster
 from dask.distributed import Client, as_completed, wait, LocalCluster, get_worker, get_client
 from dask.distributed import print as dask_print
+from enchanted_surrogates.utils.time_format import format_sec
```
DanHarJor, line 12: Just a funtion to take seconds and make a nice dd - hh:min:ss string.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -96,7 +115,18 @@ def start_cluster(self, slurm_out_dir=None):
                 self.SLURMcluster_config['job_extra_directives']+=[f'-o {slurm_out_dir}/%x.%j.out',f'-e {slurm_out_dir}/%x.%j.err']    
             print('FOR WORKER SLURM OUT, SEE:',slurm_out_dir)
             self.cluster = SLURMCluster(**self.SLURMcluster_config)
-            self.cluster.scale(self.scale_n_jobs)
+            
+            if self.submit_command:
+                self.cluster.job_cls.submit_command = self.submit_command
```
DanHarJor, line 119: This allows users to change the submit command for workers, adding some control if the dask-jobqueue mechanism is making things difficult. for example, the default is sbatch, but it can be changed to "sbatch -n 4"

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -96,7 +115,18 @@ def start_cluster(self, slurm_out_dir=None):
...
+            
+            if self.scale_n_jobs:
+                self.cluster.scale(self.scale_n_jobs)
+            elif self.scale_n_jobs_min and self.scale_n_jobs_max:
+                self.cluster.adapt(minimum=self.scale_n_jobs_min, maximum=self.scale_n_jobs_max)
+            else:
+                self.cluster.scale(1)  # Default to 1 worker if no scaling info provided
```
DanHarJor, line 126: This allows users to take advantage of dasks autoscale functionality where the user defines the min and max number of workers and dask will automatically scale up and down depending on how many tasks are submitted and left to run. I found this works quite effectivly.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -114,26 +144,34 @@ def start_cluster(self, slurm_out_dir=None):
...
+                    print(f"Worker {addr}:")
+                    print(f"  CPUs: {info['nthreads']}")
+                    print(f"  Memory: {info['memory_limit'] / 1e9:.2f} GB")
+                    print(f"  Resources: {info.get('resources', {})}\n")
+            else: 
+                warnings.warn('BLOCK UNTIL CLUSTER STARTED WAS True BUT scale_n_jobs WAS NOT DEFINED SO expected_number_of_workers COULD NOT BE DEFINED AND THE BLOCK WAS SKIPPED')
+        
+        bash_command = "scancel $(squeue -u $USER -o '%i %j' | awk '$2 ~ /dask/ {print $1}')"
+        print(f"**TOP TIP** USE THIS SLURM BASH COMMAND TO CANCEL ALL JOBS WITH dask IN THE NAME\n{bash_command}")
+
```
DanHarJor, line 173: This just moves the block feature into an if statement so it can only happen if scale_n_jobs is defined as this is needed to know how many workers to wait for. This feature is only needed in debugging and doesn't affect usual runs with block_unitl_cluster_started = False

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -219,11 +257,6 @@ def count_alive_workers(self):
                 count += 1
         return num_alive_workers
 
-    def scale(self, num_workers):
-        if self.count_alive_workers == self.scale_n_jobs:
-            self.cluster.scale(num_workers)
-        self.scale_n_jobs = num_workers
-
```
DanHarJor, line 226: Removing old function that I never got working. Now there is the dask auto scale that is working.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -242,7 +275,11 @@ def start_runs(self):
         if not os.path.exists(self.base_run_dir):
             print('MAKING BASE RUN DIR:',self.base_run_dir)
             os.makedirs(self.base_run_dir)
-
+        
+        if hasattr(self.psudo_runner,'light_pre_processing'):
+            print('PERFORMING LIGHT PRE PROCESSING FROM THE RUNNER:',self.runner_config['type'])
+            self.psudo_runner.light_pre_processing(self.base_run_dir)
+
```
DanHarJor, line 281: This allows a runner to do some light pre processing before it starts. For example in my sobol-g runner this light pre-processing is simply plotting the true function for the user to inspect. It checks if the runner has the light_pre_processing function before running it so there will be no errors for runners that do not have the function.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -254,56 +291,116 @@ def start_runs(self):
         if not self.client:
             self.start_cluster()
         print('CLUSTER STARTED')
-        all_futures = []
```
DanHarJor, line 257: I removed the need to store futures as there were indications that this might prevent scaling from being possible.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -254,56 +291,116 @@ def start_runs(self):
...
+                        
+
+                    print(f"{'_'*100}\nBATCH {self.current_batch}| [{i+1}/{num_samp_in_batch}] Futures Completed ({((i+1)/num_samp_in_batch)*100:.1f}%)","|",f"[{num_success}/{num_samp_in_batch}] Futures Succeded ({(num_success/num_samp_in_batch)*100:.1f}%)")
+                    print(f"\n TOTAL | [{completed}/{self.sampler.budget}] Futures Completed ({(completed/self.sampler.budget)*100:.1f}%)","|",f"[{all_success}/{self.sampler.budget}] Futures Succeded ({(all_success/self.sampler.budget)*100:.1f}%)")
+                    print(f"TIME PASSED: {format_sec(time.time()-start)} d - hh:mm:ss \n {'_'*100}")
+                    futures.remove(future)
+                    
+            self.current_batch += 1
+        
+        print('WALLTIME FOR ENCHANTED SURROGATES:', format_sec(time.time()-start),'d h:m:s')
```
DanHarJor, line 389: There was some restructuring so how I handle the if statements to separate the BO sampler as Aaro intended it. 

I also added the error handling here so the failed runs are placed in their own summary file.

And another summary file is made, a json with more detialed error infor on each failed run.

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -254,56 +291,116 @@ def start_runs(self):
...
         print('CLUSTER SHUTDOWN')
         self.shutdown_cluster()
+        
+        if hasattr(self.psudo_runner,'light_post_processing'):
+            print('PERFORMING LIGHT POST PROCESSING FROM THE RUNNER:',self.runner_config['type'])
+            self.psudo_runner.light_post_processing(self.base_run_dir)
 
-
-    def submit_batch(self, samples, base_run_dir=None, client=None, include_fut_to_rundir=False):
+    def submit_batch(self, samples, base_run_dir=None, client=None, include_fut_to_rundir=False, request_errors=False):
```
DanHarJor, line 402: The ability to request detailed error info for summary file creation is implimented

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -328,10 +425,13 @@ def submit_batch(self, samples, base_run_dir=None, client=None, include_fut_to_r
         run_dirs = []
         fut_to_rundir = {}
         for sample_params in samples:
-            sample_run_dir = make_run_dir(base_run_dir=base_run_dir, prepend=self.runner_config['type']) 
+            if not self.save_run_dirs:
+                sample_run_dir = make_run_dir(base_run_dir='/tmp/', prepend=self.runner_config['type']) 
+            else:
+                sample_run_dir = make_run_dir(base_run_dir=base_run_dir, prepend=self.runner_config['type'])
```
DanHarJor, line 430: This allows the user to put 

save_run_dir : False

In the config file, which puts the run directory in the /tmp/ directory which will be a different location depending on the system but will always be intended for files that will be cleaned and deleted like our run_dirs when we don't want to save them.

src/enchanted_surrogates/executors/dask_nested_executor.py
```diff
@@ -347,35 +394,40 @@ def seperate_futures(self, futures):
...
+                        else:
+                            raise RuntimeError(f'THIS FUTURE HAS BEEN SUBMITTED BUT THE RUN_DIR WAS NOT ADDED TO fut_to_rundir, future.key: {future.key}')
+                        if result == (None,None):
+                            if timeout == 0:
+                                break
+                            if not silent: print('SLEEPING TO SEE IF THE RESULT WILL BECOME AVAILABLE, sec passed:', time.time()-started)
+                            time.sleep(1)
+            else:
+                print('FUTURE NOT DONE YET, CANNOT GET RESULT NOW, SLEEPING 1 SEC AND TRYING AGAIN')
+                time.sleep(1)
```
DanHarJor, line 418: Added an if future.done() to be sure the future is actually done when getting the result.

awlishar:
> When an error happens in single_code_run, run_simulation_task catches it and returnes a different dictionary than if it had succeded. This then messes up the summary file, enchanted_dataset.csv. So I changed the executors to to make two summary files, one for succeded runs and one for failed, so they can have different headders. In run_simulation_task more error info is recorded that wasn't previously and this is optionally returned if the executor requests. Then the executor can make a file with the detailed error information for each failed run, currently this is a json file. There are other small changes, see comments in files changed

If an error happens within a runner's `single_code_run`, then isn't it more logical for the error to be handled there in the runner? The dict that  `single_code_run` returns to `run_simulation_task` should be of the same structure no matter if there was an error or not in the runner or parser. Some values of the dict can of course be empty/None/0 if the runner/parser failed at some point, but the keys shouldn't change.

DanHarJor:
Closed as it edits many of the core files, and this should be taken with more caution.

PR #144, opened 2025-11-25
added hdf5_packing into run.py

DanHarJor:


src/run.py
```diff
@@ -7,37 +7,9 @@
...
-        elif 'sampler' in config.executor:
-            config.executor['sampler_config'] = config.executor.pop('sampler')
-    if 'runner_config' not in config.executor:
-        if getattr(config, 'runner', None):
-            config.executor['runner_config'] = config.runner
-        elif 'runner' in config.executor:
-            config.executor['runner_config'] = config.executor.pop('runner')
-    print(config)
-    return config
-
```
DanHarJor, line 40: Load configuration was put into utils so it is available in the rest of enchanted surrogates when a function needs access to values in the config files. I put it in some of the new runners and samplers I believe.

src/run.py
```diff
@@ -49,9 +21,11 @@ def main(args: argparse.Namespace, config_path=None):
 
     print(enchanted_wizard)
 
+    print('GETTING EXECUTOR OBJECT')
```
DanHarJor, line 24: SOme prints so if there is a failure we know what point it got to.

src/run.py
```diff
@@ -67,8 +41,18 @@ def main(args: argparse.Namespace, config_path=None):
...
     executor.clean()
+    
+    if hasattr(args, 'general'):
+        if args.general.get('pack_data_hdf5', True):
+            # the import is placed here so that users don't need h5py installed in their enviroment to use run.py
+            from enchanted_surrogates.utils.hdf5_packer import convert_directory_to_hdf5
+            # reduce num files with hdf5
+            batch_dirs = get_batch_dirs(args.executor['base_run_dir'])
+            for batch_dir in batch_dirs:
+                convert_directory_to_hdf5(batch_dir, uuid_only=True, delete_packed=True)
```
DanHarJor, line 52: This ensures backwards compatibility. It only looks at a general namelist if it exists in the args object. 

It only imports the hdf5 packer here so if the user does not have the optional dependency then it will not break. 

It is dependant on the dask executors PR as it looks for batches.

DanHarJor (commented):
Dependant on feature/plugin_failure_detection as this has the change to save in batch dirs.

aejarvin:
Load configuration is not in utils yet, which means that merging this PR would break the code. I'll close it now. We need to resolve these PRs into incremental steps that can be tested to make sure that we do not end up with lengthy debugging procedures after the merge.

DanHarJor:
Putting load configuration in utils is apart of the PR. I don't believe anyone imports load configuration from run.py so it shouldn't break anyone's code. 

I do however agree that the smaller the PR the better. I thought this one was quite small. More automated testing is coming from the CS students I believe which will also help a lot with reducing the size of any debugging procedures.

PR #146, opened 2025-11-25
Update __init__.py

aejarvin:
I believe this should have been importlib.metadata not importlib_metadata


aejarvin:
This is included in the other pull request already.

PR #152, opened 2025-12-04
Bug/load plugins exception handling

awlishar:
The load_plugins() function looped through the plugins and tried to load them twice, and exception handling was only implemented in the first loop. The loop in the return statement failed if the plugin couldn't be loaded. Now, only the successfully loaded plugins are returned.


aejarvin (approved):
Looks good.

PR #157, opened 2025-12-11
Supervisor model

tsseppa:
### Version 1.0

Supervisor was created to ´src/enchanted_surrogates/supervisor/supervisor.py´

General architecture:
- Some tasks previously in ´run.py´ was moved to ´supervisor.py´ (path creation, executor intializing)
- Some tasks of ´Executor´ class was moved to ´supervisor.py´, see changes in ´base_executor.py´
- Sampler is no more initialized by executor. Now sampler is initialized by supervisor, which gives samples and sampler to executor (due to sampler having the function ´register_futures´ which should be accessed during execute). 
- ´local_executor.py´ was modified to work with new supervisor architecture
- Supervisor prompts user a question if ´base_run_dir´ has content inside and user wants to delete content. Content should be in current implementation deleted, as the ´enchanted_datapoints.csv´ would otherwise be appended, which breaks thing. This is due to decision to name folders with ´d{nesting_depth}_b{batch_number}_r{run_number/sample_number}´, which could possibly help with the issue Many Small Files Issue. Can be changed back if original implementation is preferred.

### Nesting:

- Supervisor now has 1...n executors and samplers:
  - Config files now have separate dicts for executors, samplers and runners. Supervisor key contains a list of run groups, where it is specified which executor, sampler and runner are used together. By adding more elements to the list, elements are nested each other.
  - All executors, samplers and runners have a name in the config files, and those names can be used as keys to reuse them in run groups.
  - Supervisor loops over all executors, runs them one by one and supplies the results of previous run as input data to the next run.
- Separate DaskNestedExecutor has been removed.

### Configuration
Some modifications made for configuration:
 - Supervisor added to config.yaml hierarchy
 - `base_run_dir` moved from executor to supervisor in config hierarchy
 - supervisor can have configuration argument `summary_datatype` with type `parquet`. This sets summary to be parquet instead of csv. ´pyarrow´ was added to dependencies
 - run_order added to supervisor, see file changes

#163 hdf5 storage also part of pull request. It has storage config attribute.

**Some changes will come from #170 **


src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+
+        # Make sure that there is nothing in base_run_dir
+        if os.path.exists(base_run_dir):
+
+            if next(os.scandir(base_run_dir), None):
+                value = input(
+                    str(os.path.abspath(base_run_dir))
+                    + "\nFolders have content. Do you want to delete data in existing folders? y/N "
+                )
```
JaaskelainenL: btw this doesn't work when running with sbatch

vanhalaj:
Refactored joblib executor to work with supervisor module and fixed unit tests for local and joblib executor. Dask executor unit tests still need to be updated.

src/enchanted_surrogates/supervisor/supervisor.py
tsseppa: i guess there is still cases where when running with sbatch there can be something in base_run_dir, so we have to find a way to 1) clean it in code OR 2) give instructions to user to delete it?

JaaskelainenL: running with sbatch does result in an output in base_run_dir and it is not empty. i think checking if stdout is a tty is a good start. then you have to make a choice to either: a) just delete it or b) just quit and prompt the user to delete it manually

JaaskelainenL:
dask nested executor should be on its own branch since it might require a lot of changes

src/enchanted_surrogates/supervisor/supervisor.py
JaaskelainenL: resolved this

docs/index.md
```diff
@@ -75,7 +75,7 @@ python enchanted-surrogates/src/run.py -cf path/to/config/file
 ```
 
 Make sure to replace `path/to/config/file` with the actual path to your configuration file. 
-The configuration file should be in YAML format and specify the runner, sampler, executor, and other parameters needed for the simulation. 
+The configuration file should be in YAML format and specify the runner, sampler, executor, supervisor and other parameters needed for the simulation. 
   
 ```yaml
 executor:
```
vanhalaj: Add run_order to supervisor and executors, samplers, runners

docs/index.md
```diff
@@ -107,15 +112,19 @@ Modules for generating samples according to some rule set.
...
+Code-specific modules for executing the code in question. Commonly paired with a code-specific parser.
 
 ### Parsers
 Code-specific modules for reading and writing files produced by the code. 
 
+### General code structure
+The supervisor is the entry point. The supervisor reads parameters and has Sampler, Executor and Runner. 
+Supervisor initializes Sampler and fetches samples from it. 
+Supervisor gives samples to executor. Executor initializes some cluster or job queue or similar. 
+A sample is sent to Runner. Runner or Parser tied to it creates files. Supervisor creates data structures
```
vanhalaj: Executor calls run simulation task and sends sample to runner

docs/supervisor.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+## Configuration of supervisor
+
+Supervisor needs `base_run_dir` defined in the configuration file.
+Example as follows:
+
+```yaml
+supervisor:
+    base_run_dir: "path/to/folder"
```
vanhalaj, line 24: Add run_order

src/enchanted_surrogates/executors/dask_executor.py
```diff
@@ -49,20 +55,23 @@ def __init__(self, *args, **kwargs):
...
-        self.scale_n_jobs = kwargs.get('scale_n_jobs', 1)
-        self.timeout = kwargs.get('timeout', 1e10)
-        self.SLURMcluster_config = kwargs.get('SLURMcluster_config')
-        self.LocalCluster_config = kwargs.get('LocalCluster_config')
-        self.block_until_cluster_started = kwargs.get('block_until_cluster_started', False)  # for debugging purposes only
+        print("INITIALISING DASK EXECUTOR")
+        # self.type = kwargs.get('type')
+        # if self.sampler_config:
+        #     self.sampler_type = self.sampler_config.pop("type")
+        #     self.sampler = import_sampler(
```
vanhalaj: Remove unused code

vanhalaj (changes requested):

vanhalaj (approved):

awlishar (commented):
Tested and works well. Amazing job!!
I will do a couple other tests tomorrow so don't merge just yet.

awlishar:
Approved for merge. :)

PR #158, opened 2025-12-11
Replace print statements with python logging module

lindste:


JaaskelainenL:
Running this on a SLURM cluster results in no ouput in the log or the SLURM output.

JaaskelainenL:
Works fine with a SLURM cluster now. LGTM

awlishar:
Looks good in general! Logging levels work as expected and "print" statements have been converted to "log" statements. However, I found one thing regarding the Dask executor. 

I am running these example config files that all use the ExampleRunner:
- **example_local.yaml**: Runs locally. The log file is created and the code writes all "log..." statements to that file. Also the log statements within the ExampleRunner are written to the log file.
- **example_dask_local/slurm.yaml**: Runs Dask clusters (local/SLURM). The general log file is created, the code writes all "log.." statements outside the cluster workers to that file. **The "log" statements within the ExampleRunner are not written anywhere/to any file.** I think the logger might not be initialized within the workers. 

However, I think this is fine. I think we can use the dask.distributed.print within the Runners instead of log. I will check with my team and get back to you.

awlishar:
> Running this on a SLURM cluster results in no ouput in the log or the SLURM output.

When testing, this was still the result for me.

DanHarJor:
We want to be able to use more executors besides dask. For example because dask workers and the simulations they run are limited to one node.

So it would be ideal if the logging could be independent of what executor we use.

awlishar:
> We want to be able to use more executors besides dask. For example because dask workers and the simulations they run are limited to one node.
> 
> So it would be ideal if the logging could be independent of what executor we use.

Logging is independent of what executor we use. You can "log" from anywhere in code as long as the logger is added to that module. Logging is pretty much a controlled way of using "print". When you develop a new executor, you just add the logger to it with the import statements, and then you can use the "log" functionality within that module as well.

DanHarJor:
> dask.distributed.print

Yes, and if we are using dask.distributed.print, this will only work if we are using the dask executor. So if we use another executor then the prints on the nodes running the simulation will not go to standard out.

So we need to make sure the logging works within single code run which is running on other nodes.

dafughes:
Still no console logs from runners when using SLURM, file logging seems to be working. When the supervisor task is complete will have to clean up some hacks like log configs for workers...

src/run.py
```diff
@@ -35,42 +40,64 @@ def load_configuration(config_path: str) -> argparse.Namespace:
...
-
 def main(args: argparse.Namespace, config_path=None):
     """
     Main function for running the simulation workflow.
 
     Args:
         args (argparse.Namespace): Namespace containing the configuration parameters.
     """
+    # Create the base run directory
+    if not os.path.exists(args.executor["base_run_dir"]):
```
vanhalaj: Might cause issues with supervisor, as it checks that base_run_dir must not exist or it must be empty.

vanhalaj (approved):

dafughes:
Now logs to console in SLURM. Some connection errors from Dask are displayed after the runs, but output from runners seems ok, maybe just Dask not shutting down cleanly?

awlishar (approved):
Tested and works, but I also get the errors in the end of the run. Probably as you wrote, some issue with shutting down the workers. I think we should still merge this one to develop. Then when both this and the supervisor have been merged, we can test how they work together and fix the errors.

src/enchanted_surrogates/executors/dask_nested_executor.py
vanhalaj, line 1: Does not exist anymore in develop

src/run.py
```diff
@@ -1,3 +1,18 @@
+# run.py
```
vanhalaj: This comment line not needed. Should module level docstring be before imports?

configs/example_bayesian_optimization.yaml
```diff
@@ -1,3 +1,5 @@
+logging: DEBUG
```
vanhalaj, line 1: Should work even if this is missing

PR #163, opened 2026-01-05
Hdf5 storage format

tsseppa:
_Under construction_

Initial commit. Plan is to have hdf5 storage format configuration in general config file, so user can create the format. Base structure should be offered (to show what could be done) and option for None as well to not have hdf5 at all.

### Current tree created into hdf5:

Näyttökuva 2026-01-06 kello 5 55 37

### Things to consider

- Metadata attributes. What should be included? (Software version, git commit, python version)?. Currently just Sampler, Executor and Runner types.
- Chunks. Depends if there is already a planned use case for hdf5 file. In some cases creating chunks fastens the process of reading hdf5 for ML purposes for example
- Sanitaizing datasets. **Currently only numeric values are saved into hdf5**. Some sanitizing should be included for True/False values. String values may not be needed in the database?
- Should this be only optional, or should the h5py package be included in dependencies?

### Todo:

- [ ] Write "How to use" guide and some examples

### DIFF
--- a/configs/example_local.yaml
+++ b/configs/example_local.yaml
@@ -10,3 +10,6 @@ runner:
   other_params: {}
 supervisor:
   base_run_dir: "data_dir/local"
+storage:
+  type: hdf5
+  path: "{base_run_dir}/runs.h5"
--- a/configs/example_local.yaml
+++ b/configs/example_local.yaml
@@ -12,4 +12,3 @@ supervisor:
   base_run_dir: "data_dir/local"
 storage:
   type: hdf5
-  path: "{base_run_dir}/runs.h5"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
     "PyYAML",
     "scipy",
     "pyarrow",
+    "h5py",
 ]
 
 [tool.setuptools]
--- a/src/enchanted_surrogates/supervisor/supervisor.py
+++ b/src/enchanted_surrogates/supervisor/supervisor.py
@@ -10,6 +10,7 @@ import sys
 import warnings
 import shutil
 from time import sleep
+import numpy as np
 import pandas as pd
 from enchanted_surrogates.utils.precise_imports import import_sampler, import_executor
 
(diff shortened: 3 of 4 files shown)


tsseppa:
# Creation of stored files

Enchanted Surrogates run results can be saved in a few ways. Some must be configured to `config.yaml` file, and some are used by default.

## CSV

Summary CSV file `enchanted_dataset.csv`containing results of runs is created **by default**. It is saved in `base_run_dir` path that has to be described in `congif.yaml`: 

```yaml
supervisor:
    base_run_dir: "path/to/location"
```

## Parquet

Instead of CSV, summary file can be saved into parquet file `enchanted_dataset.parquet`. This is created when `config.yaml` is configured as follows and it replaces the summary .csv file:

```yaml
supervisor:
    summary_datatype: "parquet"
``` 

## HDF5

In addition, data can be saved to hdf5 which is structured storage format. Currently the file structure is as follows:

```text
- runs.h5/
    - data/
        - aggregated/
            - columns
            - values
        - runs/
            - 0_0/
                - columns
                - values
            - 0_1/
                - columns
                - values 
            - ...
            - 1_0/
                - columns
                - values
            - ...
    - metadata/
        - attributes
```



In `config.yaml` hdf5 has to be declared as follows: 

```yaml
storage:
    type: "hdf5"
```

Currently there are no other storage types, so as long as `type` parameter is not empty, the `runs.h5` file will be created

# Use of stored files

## CSV

CSV file can be opened in LibreOffice Calc or other spreadsheet application. Columns used are:
- `output`: float; result value
- `success`: bool; state of result
- `c1`: float; simulation parameter
- `c2`: float; simulation parameter
- `run_dir`: path; location of run for this result

Number of rows corresponds to number of runs conducted.

## Parquet

Parquet file is organized similar to csv described above. It is possible to be accessed via script as below:

```python
import pandas as pd

# Load summary dataset
df = pd.read_parquet("path/to/enchanted_dataset.parquet")

# Inspect contents
print(df.head())
print(df.columns)

# Example for ML purposes
# Feature matrix (simulation inputs)
X = df[["c1", "c2"]]

# Target vector (simulation output)
y = df["output"]
```

## HDF5

File structure is described above. Below is how to inspect the values of hdf5 file with usage of h5py package:

```python
import h5py

with h5py.File("path/to/runs.h5", "r") as f:
    # Top-level groups
    print(list(f.keys()))  # ['data', 'metadata']

    # Data subgroups
    print(list(f["data"].keys()))  # ['aggregated', 'runs']

    # Available run IDs
    run_ids = list(f["data"]["runs"].keys())
    print(run_ids)
```

A very basic example for ML purposes:

```python
import h5py
import numpy as np

with h5py.File("path/to/runs.h5", "r") as f:
    columns = f["data"]["aggregated"]["columns"][:].astype(str) # instance of np.array()
    values = f["data"]["aggregated"]["values"][:] # instance of np.array()

# Extract features and target
c1_idx = columns.tolist().index("c1")
c2_idx = columns.tolist().index("c2")
out_idx = columns.tolist().index("output")

X = values[:, [c1_idx, c2_idx]]
y = values[:, out_idx]
```

dafughes (approved):
Looks good!

PR #164, opened 2026-01-08
Feature/automated documentation from docstrings

shcera:
This branch introduces an automated documentation pipeline for the project, centered around docstrings and MkDocs-based static site generation. 
All samplers were updated to include structured docstrings. Each sampler now documents: A concise overview describing the purpose and behavior of the sampler, input parameters and configuration options as they appear in the configuration files.
Expected outputs, including return types and data formats, assumptions and implementation notes, such as distribution choices or limitations. Similarly, docstrings for all existing executors were extended and unified.
All documentation is written directly in the docstring of the concrete class, ensuring that the documentation lives next to the implementation.

Automatic Documentation Generation with MkDocs
To turn these docstrings into user-facing documentation, automatic documentation generation was set up using MkDocs together with mkdocstrings.
A dedicated mkdocs.yml configuration file was added to the repository. This file defines:
The site metadata (site name, description, repository link).
The documentation theme and styling options.
The navigation structure of the documentation site.
The mkdocstrings plugin configuration, which enables automatic extraction of documentation from Python docstrings.
The navigation section of mkdocs.yml explicitly defines how the documentation is organized. Typically, this includes:
A main landing page (index.md) with a project overview.
Separate sections for samplers and executors.
Individual pages that reference Python modules or classes using mkdocstrings, which pulls in the corresponding docstrings automatically.
When MkDocs is built, it scans the documentation files in the docs/ directory. Whenever a page contains a mkdocstrings directive pointing to a Python module or class, MkDocs imports that object, reads its docstring, and renders it into structured HTML according to the configured style. This means that updating a docstring in the source code is sufficient to update the published documentation.

Continuous Deployment with GitHub Pages
To make the documentation available automatically, a GitHub Actions workflow was created. This workflow builds and deploys the documentation site to GitHub Pages whenever changes are pushed to the repository. 
The workflow performs the following steps:
Checks out the repository source code.
Sets up a Python environment compatible with the project.
Installs MkDocs and all required documentation dependencies (including mkdocstrings).
Runs mkdocs build to generate a static documentation site.
Deploys the generated site to the gh-pages branch (or the GitHub Pages deployment target), making it publicly available.
This ensures that the documentation is always up to date with the latest changes in the codebase and requires no manual intervention after pushing changes.

Required Project Settings
To enable this setup, the following project settings must be configured in GitHub:
The source for GitHub Pages should be set to GitHub Actions.
The workflow file must have permission to write to the repository.


src/enchanted_surrogates/samplers/array_sampler.py
```diff
@@ -1,45 +1,46 @@
...
-    The samples are directly specified in 'bounds' in the config file.
-
-    Attributes:
-        bounds (list of list of float or int): The bounds of each parameter.
-        num_samples (list of int): The number of samples for each parameter.
-        parameters (list of str): The names of the parameters.
-        budget (int): The total number of parameter combinations.
-        num_initial_points (int): The number of initial points in the sample space.
-        samples (list of list of float or int): The generated parameter combinations.
-        current_index (int): The index of the current parameter combination.
```
tsseppa, line 21: Should Attributes and Methods still be written the file documentation, or is deleting them according to some new documentation convention? Other executors are storing the methods, but not all are storing the attributes but just arguments. Also, all other executors have docsring inside the class, should this be also so? If these are thought through, please elaborate made design choices.

tsseppa (changes requested):
Check comment. Made into ArraySampler. If intentional design, ready to merge. If there is no reason to treat ArraySampler differently than other Sampler documentation, make changes.

.github/workflows/documentation.yml
tsseppa, line 1: We discussed about this lately in dailies, we should somehow test that this works before having it merged to main. Easiest way would be to add this very branch to 

```yaml
on:
  push:
    branches:
      - main
      - feature/automated-documentation-from-docstrings
```

and then make sure deployent would be done only when the branch is main

```yaml
deploy:
  if: github.ref == 'refs/heads/main'
  needs: build
  runs-on: ubuntu-latest

```

I am not sure but I think this could be enough to have it in CI/CD loop and see if there are any errors.

tsseppa (changes requested):

tsseppa (approved):

tsseppa (approved):

awlishar (approved):
Well done! Feel free to merge to 'develop'.

PR #165, opened 2026-01-13
Feature/enc 122 works for dask nested executor

vanhalaj:
_Note: this branch is merged into supervisor branch, not develop_

Supervisor now has logic for nesting executors and samplers. 

### Changes:
- Supervisor now has 1...n executors and samplers.
  - Config files now have separate dicts for executors, samplers and runners. Supervisor key contains a list of run groups, where it is specified which executor, sampler and runner are used together. By adding more elements to the list, elements are nested each other.
  - All executors, samplers and runners have a name in the config files, and those names can be used as keys to reuse them in run groups.
  - Supervisor loops over all executors, runs them one by one and supplies the results of previous run as input data to the next run.
- Separate DaskNestedExecutor has been removed.

### Notes:
- Workflow tests and all example configs still need to be updated to use the new config style.
  - This causes pylint pipeline to fail for now
- Currently some nesting-related code is run even in single-executor configurations. This does not cause issues, but causes extra computation steps.

These will be fixed separately. This PR is only for nesting logic.


src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -70,20 +77,46 @@ def start(self):
...
-                for i in range(len(samples))
-            ]
-            # Call executor with folder path and samples in tuple
-            self.executor.execute(zip(run_dirs, samples), self.sampler)
-
-        self.wait_all_processes()
-        enchanted_dataset = self.create_dataset()
+        rows = [{}]
+        for depth, group in enumerate(self.groups):
+            next_rows = []
```
tsseppa, line 82: next_rows could be merged with rows?

src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -176,23 +210,24 @@ def all_processes_done(self):
...
+                        return False
 
         return True
 
-    def wait_all_processes(self):
+    def wait_all_processes(self, filter=None):
         """
         Waits in while loop until all simulations are done. Loop is broken
         when all_processes_done returns true. Checks condition once in
         second.
```
tsseppa, line 229: Missing docstring for filter

tsseppa (changes requested):

src/enchanted_surrogates/supervisor/supervisor.py
vanhalaj: Fixed

src/enchanted_surrogates/supervisor/supervisor.py
vanhalaj: Can't be merged, because rows is accessed in every batch and used to create unified table of parameters from previous run and parameters received from get_next_samples. So rows can only be updated in the end of outer loop.

This should be compatible with better "update summary files in batches" -implementation in #170

tsseppa (approved):

PR #170, opened 2026-01-20
Feature/enc 90 seamless sampling continuation

dafughes:
Supervisor now allows for seamless sampling in case a sampler or the supervisor crashes. Checkpoints are written after each batch. This also works with nesting. Tests are also included

Run workflow tests with `pytest .\workflow_tests\automated_tests_no_HPC\test_sampling_continuation.py`

### fresh

Normal run and the default

### resume

- Resume an interrupted previous run: 
  - Set `run_mode: resume` under `Supervisor` in the config file, no other changes are needed. 
  - Supervisor keeps track of the run state (how many batches sampled, which nesting depth, and how many samples have been submitted) and using this data restores the run from where it left off and continues from there.

- Resume a previous run that was completed but increase the budget:
  - Set run  mode to resume and also increase the budget of the sampler. From Supervisor point of view, this is same as if the budget always was that much and the run was just interrupted.
  - Eg. increasing budget from 50 to 60 and re-running generates 10 new samples.

### extend

- Extend a previous run:
  - Set `run_mode: extend` under `Supervisor` in the config file
  - Now setting sampler budget to 10 means that 10 new samples are created.

These work for nested sampling as well.


tsseppa:
First implementation done. Requires reviewing and testing.

Next steps:
- Unit tests
- System testing

src/enchanted_surrogates/samplers/random_sampler.py
```diff
@@ -24,7 +24,7 @@ def get_next_samples(self) -> list[dict]:
 
     def register_future(self, future):
         """ Doesn't matter for random sampler TODO: Probably? """
-        return None
+        return None
```
vanhalaj: Remove added whitespace

configs/example_local.yaml
```diff
@@ -10,5 +10,6 @@ runner:
   other_params: {}
 supervisor:
   base_run_dir: "data_dir/local"
+  run_mode: "fresh" # "resume" / "extend"
```
vanhalaj, line 19: I think it would be good if sampling could be continued without editing config file. For example if the program has crashed, just running it again with no changes would be able to continue. Alternatively, maybe a command line argument like "--continue" could be given when running the program

src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -54,6 +55,23 @@ def __init__(self, args, config_path=None):
...
+                    previous_run_data["batch_number"] +1
+                )
+                self.batch_number = previous_run_data["batch_number"]
+            else:
+                raise RuntimeError(
+                    "Tried to continue from previous sampling but no "
+                    " enchanted_run.yaml was found"
+                )
+
         self.create_base_run_dir(self.base_run_dir, config_path)
```
JaaskelainenL: create_base_run_dir is still ran and created even if last run is extended or resumed. it will halt the program due to existing files. it should either not run or delete the last incomplete batch (prompt user maybe?).

configs/example_local.yaml
JaaskelainenL: i think we could use a different config for testing for bigger runs. maybe just copy example_local but add batch_size to the sampler and make the budget big

src/enchanted_surrogates/supervisor/supervisor.py
JaaskelainenL: all ok now

tsseppa:
Update 10.2. Many small files issue was merged here, it was approved by client in its own branch.

awlishar:
Could you please describe run modes "resume" and "extend"?
"fresh" works as expected.

## Testing "resume"

I started with the following config file:

```
logging: DEBUG

executors:
  e1:
    type: LocalExecutor
samplers:
  s1:
    type: RandomSampler
    bounds: [[0.0, 1.0], [0.0, 10.0]]
    budget: 12
    batch_size: 4
    parameters: ['c1', 'c2']
runners:
  r1:
    type: ExampleRunner
    sleep_sec: 5.0

supervisor:
  base_run_dir: "data_dir/local"
  run_mode: "fresh" # "resume" / "extend"
  save_files: "all" # or "custom" or "none"
  save_files_list:
    - enchanted_dataset.csv
    - example_local.csv
  run_order:
  - executor: e1
    sampler: s1
    runner: r1

storage:
  type: csv
```

I stared the ES run from the command line:

`python src/run.py -cf configs/example_local.yaml `

Since I defined sleep_sec to 5 seconds in the ExampleRunner, I can easily follow the samples being created and added to the enchanted_dataset.csv. The samples are correctly added to the enchanted_dataset.csv file as every batch is finished. The budget is 12 but I "Ctrl+C" in the command line to "crash" the run after 8 samples. Then I change run_mode to "resume" and start ES again in the same way. The 4 remaining samples as correctly created.

However, they all samples generated from "resume" have index "0" in the enchanted_dataset.csv.


## Testing "extend"
I would assume that the "extend" option makes it possible to append to an existing dataset.

```
logging: DEBUG

executors:
  e1:
    type: LocalExecutor
samplers:
  s1:
    type: RandomSampler
    bounds: [[0.0, 1.0], [0.0, 10.0]]
    budget: 10
    batch_size: 5
    parameters: ['c1', 'c2']
runners:
  r1:
    type: ExampleRunner
    other_params: {}

supervisor:
  base_run_dir: "data_dir/local"
  run_mode: "fresh" # "resume" / "extend"
  save_files: "all" # or "custom" or "none"
  save_files_list:
    - enchanted_dataset.csv
    - example_local.csv
  run_order:
  - executor: e1
    sampler: s1
    runner: r1

storage:
  type: hdf5
```

Correctly creates 2 batches of 5 samples each.
Then I used the same config file but changed only

```
  run_mode: "extend"
```

Ran the simulation again and nothing happened. 
Then I increased the number of samples (budget).

```
samplers:
  s1:
    type: RandomSampler
    bounds: [[0.0, 1.0], [0.0, 10.0]]
    budget: 12
    batch_size: 5
    parameters: ['c1', 'c2']

```

Now this added one new batch with 50 samples to the data set. I was expecting it to perhaps extend the data set with 2 samples since I increased the budget with 2. 

Perhaps I am using the "extend" option wrong?

vanhalaj:
@awlishar I now added description of the functionality to the PR description. There was an issue that resume and continue did the completely same thing which has now been fixed.

Also the index column containing all zeroes was removed, it had been added by accident when transitioning logic from executors to supervisor (that column is added in by default in the to_csv function).

vanhalaj:
Small changes: Fixed merge conflict with develop branch and reworked workflow tests to use yaml files

JaaskelainenL:
changes look good :+1:

awlishar (approved):
Tested and works! 👍

PR #172, opened 2026-01-24
Nix support

### LINKED ISSUE #171: hard to run on nixos
I had some trouble running this on nixos because of numpy. Maybe having a script which automates the setup with nix-shell could be nice?
Nixos is pretty popular so there might be others in the future with this issue :)

JaaskelainenL:
i had some trouble running this with nixos and it would be nice to have an automated nix-shell script so others wouldn't have to do it again. 

nixos is getting pretty popular so others might face this issue.

this also provides value to people using nix on other platforms than nixos to manage development environments

also fixes #171

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,43 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+let
+  python = pkgs.python313;
+in
+pkgs.mkShell {
+  packages = with pkgs; [
+    stdenv
+
+    python
+    python.pkgs.pip
+    python.pkgs.virtualenv
+
+    gcc
+    gfortran
+    openblas
+    pkg-config
+    cmake
+    ninja
+    patchelf
+  ];
+
+  OPENBLAS = pkgs.openblas;
+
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -0,0 +1,43 @@
+{ pkgs ? import <nixpkgs> {} }:
```
vanhalaj, line 1: Maybe this script could be under some new subdirectory, such as '/scripts'? Because in the future, if similar scripts are needed eg. for other OS's, they don't fill the project root.

vanhalaj (dismissed):

shell.nix
JaaskelainenL: this is a good idea but for now i think keeping it in root could be fine because it is used to set up a dev environment and it works cross OS.

JaaskelainenL:
ready for client review

awlishar (approved):
I cannot test this, but since it does not affect any other part of the code and is nice to have, let's just merge it. :)

PR #175, opened 2026-01-29
Ascii art

Husseinhamsa:


JaaskelainenL:
LGTM

DanHarJor:
I appreciate the effort, It is good you make it smaller. It is a little hard to read the text and the wizard is a little hard to make out. I think we should keep what we have for now.

It would be better to have something that fits more neatly in a small terminal window. But this isn't priority. Feel free to focus on other tasks.

awlishar:
How about putting both words on the same line and simplifying the wizard to for example only a wizard hat?

DanHarJor:
Sounds like a good idea.

JaaskelainenL:
can you maybe make a function to choose a random one and then call that in main?

DanHarJor (dismissed):
These are great, thanks. I think version 7 is the best. I vote we stick to printing that one and leave the rest for inspiration incase others want to have a go at making one.

DanHarJor (approved):
Looks great, thanks, lets merge

PR #177, opened 2026-02-03
feat: possible to delete automatically unwanted files in base_run_dir

tsseppa:
with config file it is now possible to delete unwanted files from base_run_dir

in config.yaml
```yaml
supervisor:
  save_files: "all" # or "custom" or "none"
  # if using custom, only described files are saved
  save_files_list:
    - file.txt
    - file2.txt 

```

**Note: `enchanted_dataset.csv` and `runs.h5` are saved by default.**


JaaskelainenL:
looks good!

DanHarJor:
This looks great.

It would be better if the unwanted files were deleted as soon as the run is finished and the enchanted_datapoint.csv is read. Some people are only allowed a few hundred thousand files on their system and want millions of data points. So we need to be able to clean as we go. 

In general the repo is looking great, thanks a lot.

src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -477,3 +482,31 @@ def create_hdf5(self, enchanted_dataset: pd.DataFrame):
                     run_group.sampler.__class__.__name__
                 )
                 meta_run_group.attrs["runner"] = str(run_group.runner.get("type"))
+
+    def save_files(self, argument: str):
```
vanhalaj: Should this function be renamed, as the point is to delete unwanted files instead of saving wanted files?

tsseppa: I guess I should go for something like delete_unwanted_files

tsseppa:
@DanHarJor It should now work so that it cleans as it goes.

DanHarJor:
I am happy for you to merge it :)

PR #181, opened 2026-02-09
chore: added DS_Store to .gitignore

tsseppa:
MacOS creates .DS_store file to folders which have been explored by user. This should be in .gitignore for convenience.


vanhalaj (approved):

JaaskelainenL (approved):
I don't know how many other mac developers there are on this team but this might help other developers in the future. At least we won't accidentally get DS_Stores in PRs

PR #182, opened 2026-02-10
feature: added default base_run_dir for local use

tsseppa:
Added for local use "base_run_dir" as base_run_dir location. Cluster usage requires base_run_dir defined in config.

### DIFF
--- a/src/enchanted_surrogates/supervisor/supervisor.py
+++ b/src/enchanted_surrogates/supervisor/supervisor.py
@@ -67,7 +67,10 @@ class Supervisor:
         self.base_run_dir = args.supervisor.get("base_run_dir")
 
         if self.base_run_dir is None:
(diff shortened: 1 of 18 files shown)


src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -67,7 +67,10 @@ def __init__(self, args, config_path=None):
         self.base_run_dir = args.supervisor.get("base_run_dir")
 
         if self.base_run_dir is None:
-            raise ValueError("base_run_dir is not set in the provided configuration")
+            if sys.stdout.isatty():
+                self.base_run_dir = "base_run_dir"
```
vanhalaj, line 77: Could be logged that it has not been set and default value is used instead.

PR #184, opened 2026-02-10
Active Learning

JaaskelainenL:
Active learning sampler

This PR also moves future registering to the futures.

The model in ActiveLearningSampler is hardcoded currently but the query strategy is configurable


src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -15,6 +15,8 @@
 import h5py
 import numpy as np
 import pandas as pd
+from skactiveml.regressor import NICKernelRegressor, SklearnRegressor
```
DanHarJor: sklearn should not be imported in the supervisor. All sklearn related code should be in the active_learning_sampler you are making.

pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "scipy",
     "pyarrow",
     "h5py",
+    "scikit-activeml"
```
DanHarJor: this should be an optional dependancy that only needs to be installed if the user wants to use your active_learning_sampler

src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -106,6 +108,9 @@ def start(self):
 
         print("Starting runs...")
 
+        nick_regressor = NICKernelRegressor()
+        model = SklearnRegressor(BaggingRegressor(nick_regressor, n_estimators=5))
```
DanHarJor: the model should be defined in the config file that sets up the sampler. there should be no sklearn code in the supervisor but only in the sampler. 

The supervisor only serves to get samples from the sampler, ask the executor to run them, monitor the enchanted_datapoint.csv file to get the output of each sample. Then it needs to somehow pass this information to the sampler for the active learning

It can pass it as a python object

It can simply make the sumary csv file and let the sampler read it from the csv file

or anything else you can imagine and think is best. 

Once the sampler has these samples that have already been ran then get_next_samples() should train a ml model. Compute the acquisition function for a pool of unlabeled samples and select the next samples to be ran and return them.

src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -123,7 +128,7 @@ def start(self):
                     last_complete_dataset = self.read_summary(
                         "last_complete_enchanted_dataset"
                     )
-
+            group.sampler.set_model(model)
```
DanHarJor: model should be defined in the sampler configs and in the sampler itself.

src/enchanted_surrogates/supervisor/supervisor.py
DanHarJor, line 1: I don't see anywhere that you call sampler.register_future() to update the X_obvs and the y_obvs with the new data. 

This should be done in the while loop with each row in df_batch being a 'future'.

The way you have set this up, you need to make sure the runners single_code_run returns an output dict with 'y' as one of the keys and the value is the scalar output being considered for the active learning.

src/enchanted_surrogates/supervisor/supervisor.py
JaaskelainenL: From slack messages and from the last retro, I got the idea that the model should be on the supervisor. It originally fully lived within the active_learning_sampler but got asked to move to the supervisor. Should I move it back?

src/enchanted_surrogates/supervisor/supervisor.py
JaaskelainenL: This PR is still a draft so that hasn't been implemented yet, but it will be after we get the hang of this task better. We still haven't written tests and it's quite unclear how this should work in the big picture. Originally the learning happened within the sampler but it got moved after a request.

tsseppa:
Created tests, but due to uncertainity of the query strategy (always some unpredictability) tests are now passing approximately 7/10 times. GreedySamplingTarget (if I got it correctly) should pick x that has predicted y value that is furthest away from the already acquired y values. This does not work on the sin curve when QS picked only 1 sample (it goes for maximum y insted), but works 7/10 times with 3 QS picking three samples. Base setting was using 5 uniformly distributed x values between 0.0 < x < 1.6 and then QS choosing.

I should maybe add some probabilistic stuff so there would be 10-20 iterations and then it is checked if 60-70% of cases are successful.

DanHarJor:
> Created tests, but due to uncertainity of the query strategy (always some unpredictability) tests are now passing approximately 7/10 times. GreedySamplingTarget (if I got it correctly) should pick x that has predicted y value that is furthest away from the already acquired y values. This does not work on the sin curve when QS picked only 1 sample (it goes for maximum y insted), but works 7/10 times with 3 QS picking three samples. Base setting was using 5 uniformly distributed x values between 0.0 < x < 1.6 and then QS choosing.
> 
> I should maybe add some probabilistic stuff so there would be 10-20 iterations and then it is checked if 60-70% of cases are successful.



Hello, glad it is working. The tests shpuld be designed to make sure there are no coding errors. The query strategy is more of a user choice and it is up to them to use one that works. 

The query strategies are trying to pick the new point that provides the most information towards a goal.

In general greedy sampling query strategy will spread out the points and this is similar to picking the point with the largest distance. But there are many greedy sampling strategies and they are never exactly picking the point with largest distance. If that was the actual goal then the query strategy would be just to pick the point with the largest distance from current points. 

Dont try and test the query strategy in that way. Just test that everything runs smoothly.

tests/executors/test_joblib_executor.py
```diff
@@ -1,31 +0,0 @@
...
-
-    executor = JoblibExecutor(
-        runner_config={"type": "mock_runner"}
-    )
-
-    executor.execute(executor_input, sampler)
-
-    # Executor should not directly call get next samples
-    assert sampler.get_next_samples.call_count == 0
-    assert sampler.register_futures.call_count == 1
```
vanhalaj, line 30: We can keep this test but just remove this assert for register_futures call count

tests/executors/test_local_executor.py
```diff
@@ -1,24 +0,0 @@
...
-
-    executor = LocalExecutor(
-        runner_config={"type": "mock"}
-    )
-
-    executor.execute(executor_input, sampler)
-
-    # Executor should not directly call get next samples
-    assert sampler.get_next_samples.call_count == 0
-    assert sampler.register_future.call_count == 2
```
vanhalaj, line 23: Same for this

tests/executors/test_joblib_executor.py
JaaskelainenL: readded

tests/executors/test_local_executor.py
JaaskelainenL: readded

src/enchanted_surrogates/samplers/active_learning_sampler.py
```diff
@@ -0,0 +1,197 @@
...
+    - Warmup behavior: while `self.submitted < self.warmup` or no
+      observations exist, `get_next_samples()` returns random fallback
+      samples.
+    - The surrogate model and query strategy are dynamically
+      instantiated; these should be made configurable (for example via
+      external config) if different regressors or ensemble parameters are
+      required.
+    """
+
+    BATCH_SAMPLE_SIZE = 1
```
vanhalaj: Unused?

vanhalaj:
Using this config file:
```
logging: DEBUG

executors:
  e1:
    type: LocalExecutor
samplers:
  s1:
    type: ActiveLearningSampler
    bounds: [[0.0, 1.0], [0.0, 10.0]]
    budget: 64
    batch_size: 4
    parameters: ['c1', 'c2']
    query_strategy: GreedySamplingTarget
runners:
  r1:
    type: ExampleRunner
    other_params: {}

supervisor:
  base_run_dir: "data_dir/active_ml"
  run_order:
  - executor: e1
    sampler: s1
    runner: r1
```

Run fails: 
```
active_learning_sampler.py, line 192, in register_future
    params, y = future
    ^^^^^^^^^
ValueError: too many values to unpack (expected 2)
```

It seems that `params` has the value `output`. Maybe futures are parsing the wrong row from futures, row with names instead of row with values?

vanhalaj:
Looks good!

aejarvin (approved):
The implementation looks good. I cloned and installed the branch and tried the example_active.yaml without any problems. A future development suggestion is to add some kind of analysis features to the example to confirm that the active sampling acquisition works as intended. As an example, I used the plotting features in lampe to generate diagnostic figures between batches within the Bayesian optimization. This does not need to be added here.

DanHarJor (commented):
I don't see any obvious errors or defects. If the tests are running fine then lets merge.

Side note:
The example_active.yaml should run but it a little strange to have a different active sampler for each runner. The example_active.yaml for now should just have one sampler and one runner.

Maybe in the next sprint we will tackle active learning with nested pipelines but for now non nested is fine, just one sampler and one runner for active examples. Although since this is not a bug and it will run there is no need to edit before merging.

PR #185, opened 2026-02-10
fix: Plugins and modules are only imported once

### LINKED ISSUE #176: bug: Plugins are re-loaded at every import
Every time import_sampler, import_executor and import_runner are called, the plugins are imported again. I should be enough to import them once and then just access them.

vanhalaj:
Fixes #176 

## Changes:

- Added helper function "cached_import" that is now used by all precise import functions:
  - For plugins: On first call, import all plugins and store in memory
  - For library modules: Check if already loaded, or otherwise import and store in memory
- created unit tests for verifying imports are only done once and for importing executors, samplers and runners

### DIFF
--- a/src/enchanted_surrogates/utils/precise_imports.py
+++ b/src/enchanted_surrogates/utils/precise_imports.py
@@ -1,3 +1,7 @@
+"""
+TODO
+"""
+
 import re
 import importlib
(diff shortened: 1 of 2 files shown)


tsseppa:
Looks good to me!

awlishar (approved):
Great!

PR #186, opened 2026-02-10
Pre commit hook

Husseinhamsa:
Update (Mar 5th):
Hooks are moved to .githooks because files under .git cannot be added to the repo. Therefore I propose a solution where developers are asked to enable the custom hooks (maybe in README when that gets updated) by running:
`git config core.hooksPath .githooks`

Also using a bash script seems like a better solution than the python pre-commit package, since that requires an unnecessary package and manual git hook installation is still required. Maybe in the future we can add yamllint or pytest too
- @JaaskelainenL


data_dir/local/example_local.yaml
vanhalaj, line 1: This file should not be included

awlishar:
Please remove the `data_dir/local/example_local.yaml` from the PR and add a description for the functionality. :)

laurisaksi (dismissed):
You should not remove the file `configs/example_local.yaml` it was meant not to include the `data_dir/local/example_local.yaml` into the PR

JaaskelainenL:
I picked up this branch and propose a new solution

JaaskelainenL:
* new implementation, same branch

awlishar:
I tested by making a new branch based on this one (test-pre-commit-hook) and made some test commits but I don't think anything was triggered. 

image

JaaskelainenL:
Did you remember to set the hook path to .githooks using
`git config core.hooksPath .githooks`

Git doesn't allow pushing stuff within the .git folder to the branch so getting a pre-commit hook to work requires some fiddling

awlishar:
Ah nope sorry I didn't do that. It works :) Is there a way to force a commit even if it contains linting errors?

JaaskelainenL:
you can use git commit --no-verify to skip it

awlishar (approved):
Great!

PR #187, opened 2026-02-16
test: Running workflow tests from config files

vanhalaj:
Workflow tests are now run from test config files instead of being hard coded into the test cases.

## Changes

- Added GitHub Actions workflow to run workflow tests for pull requests and when merged/pushed to main or develop
- Updated automated (no HPC) workflow tests to use test config files, so they are now faster to make and also manual running is now easy
- Fixed issue where logger initialization in `run.py` was causing dask-localslurm tests to fail. Also logger singleton can now be reseted (which is automatically done in every test cleanup)
- Fixed bug where dask cluster shutdown can be called multiple times when reusing them

### DIFF
--- a/.github/workflows/pytest_workflow.yml
+++ b/.github/workflows/pytest_workflow.yml
@@ -3,9 +3,9 @@ name: Workflow testing with Pytest
 on: 
   workflow_call:
   push:
-    branches: [main]
+    branches: [main, develop]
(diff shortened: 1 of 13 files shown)


lindste (approved):
LGTM 👍

awlishar (approved):
Looks good!

PR #189, opened 2026-02-19
feature: creating default base_run_dir in local use if not configured

tsseppa:
Creating default base_run_dir for local use if there is not one described in configuration file.


vanhalaj (approved):
Looks good!

awlishar (approved):
Good!

PR #190, opened 2026-02-19
Feature/enc 129 automated documentation from plugins

shcera:
This branch introduces an example of automatic documentation generation using the Mishka plugin, along with minor improvements to the core documentation generation workflow.
Mishka Plugin Integration
The plugin documentation is pulled from a private repository and integrated into the Python path using the following workflow:
- name: Checkout private docs from plugin Mishka
  uses: actions/checkout@v4
  with:
    repository: DIGIfusion/enchanted-plugin-mishka
    token: ${{ secrets.MISHKA_TOKEN }}
    path: plugins_docs/mishka_plugin
- name: Add private plugin to PYTHONPATH
  run: echo "PYTHONPATH=$(pwd)/plugins_docs/mishka_plugin/src" >> $GITHUB_ENV

After this, the plugin module can be imported and documented on the generated pages. For example, in mishka_plugin.md:
::: enchanted_plugin_mishka
    options:
      show_source: false
      show_submodules: true
      members: true
      show_root_heading: true
This ensures the plugin's classes, functions, and submodules are automatically included in the documentation.

Other update for automated documentation: docs/gen_ref_pages.py has been updated to remove the need to import modules directly in the docs folder. If the source code contains docstrings, they are now imported automatically, allowing full documentation to be written directly in docstrings. Separate Markdown files can still be used if needed. Important: The first docstring should appear before imports, e.g., before import numpy, to be picked up correctly.

The main focus of this pull request is documentation generation from the Mishka plugin, but the implementation could be extensible to other plugins if this way to implement documentation pulling from plugins is satisfiable.


awlishar (approved):
Good!

PR #193, opened 2026-03-05
Feature/enc 47 performance increase from io operations

lindste:


JaaskelainenL:
LGTM

awlishar (approved):
Tested with both defining the path as a env variable and directly as a string - works!

tsseppa:
About resolving conflicts, with current implementation

```yaml

base_run_dir
    -data
        -d0_b0_r0
        etc.
    -configs
    -logs

```

Currently files are saved to self.data_dir location, which is base_run_dir/data
Keep that in mind when resolving the conflicts!

vanhalaj:
Does this work together with #177 ?

PR #194, opened 2026-03-05
feat: added data, configs, logs subfolders to base_run_dir

tsseppa:
Data, logs and configs have now their own subfolders. Structure is as follows:

```yaml
base_run_dir
  data
    b0_d0_r0
    ...
  logs
  config
  enchanted_datapoints.csv
  runs.h5

```

SLURM was also modified to have logs placed in the logs file.

### DIFF
--- a/src/enchanted_surrogates/supervisor/supervisor.py
+++ b/src/enchanted_surrogates/supervisor/supervisor.py
@@ -319,7 +319,11 @@ class Supervisor:
 
         # Create base run dir and data dir inside it
         os.makedirs(base_run_dir, exist_ok=True)
+<<<<<<< HEAD
         os.makedirs(self.data_dir, exist_ok=True)
+=======
+        os.makedirs(os.path.join(base_run_dir, "data"), exist_ok=True)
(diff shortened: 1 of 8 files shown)


vanhalaj:
I think the workflow tests need minor changes to support this. Should be enough to make the change to `get_run_dir_count` in `test_utils.py` which is included in develop since #187

src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -110,8 +110,9 @@ def start(self):
                         expanded.append(merged)
 
                 # Create run directories named by depth, batch and sample numbers
+                data_dir = os.path.join(self.base_run_dir, "data")
```
vanhalaj: `os.path.join(self.base_run_dir, "data")` is repeated now every time the data path is needed. Would it be better if it would instead be stored as class variable in init?

Looks good otherwise.

tsseppa: I think that is quite reasonable

vanhalaj:
Looks good!

awlishar (approved):
Tested and works!

PR #208, opened 2026-03-14
friedman benchmark for active learning and nested active learning

DanHarJor:
I added the friedman function as a benchmark function for active learning. I also added some unit tests. The friedman_active.yaml and friedman_random.yaml files will not run because the necessary infrastructure is currently not present in the supervisor. But it is necessary that they are possible to run.

It is likely that active learning will only outperform random sampling for small number of samples, like 20 or 50 samples.

I am keeping my eyes open for a good benchmark high D problem where active learning strongly outperforms random sampling.


awlishar:
The sampler and runner types should be defined with the class names not the file names in the configs. RandomSampler, not random_sampler.

What is the purpose of example_nested_friedman_random.yaml in comparison to example_nested_friedman.yaml?

configs/example_sequential_friedman_random.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+# get_next_samples for this sampler should only be called once after all the runners have ran. 
+# the supervisor is currently not capable of this, please make it possible
+sampler:
+  type: random_sampler
+  bounds: [[0.0, 1.0], [0.0, 1.0], [0.0, 1.0], [0.0, 1.0], [0.0, 1.0]]
+  budget: 30
+  batch_size: 1
+  parameters: ['x1','x2','x3','x4','x5']
+
```
JaaskelainenL, line 20: Did you mean this to be split like on example_nested_friedman.yaml? Samplers have to be defined as a list

PR #213, opened 2026-03-19
feat: Multi-runner Sequential Execution

JaaskelainenL:
Multi-Runner Sequential is implemented into the supervisor, in a way which allows the code to run either simple, sequential, nested or nested+sequential sampling.

#### Config files

- Sequential runners can be used by specifying multiple runners and executors in one run group, f.ex:
```yaml
run_order:
  - sampler: code1_sampler
    executor:
      - code1_executor
      - code2_executor
    runner:
      - code1_runner
      - code2_runner
```
- So, executor and runner can have values that are either lists or single values as before

#### Supervisor

- Added new innermost loop to `start` function, which loops over all sequential executor+runner combinations of the current run group
- `group` renamed to `nested_group` to avoid confusion with sequential runners and executors
- Run dirs now have the format `d#_b#_s#_r#`
- Refactored import functionality (moved from Supervisor to nested_imports.py)
- Fixed small bug where seamless sampling might not have been working properly in nested runs, if execution stops before the final nesting depth

#### Misc

- `runner_config` is now given as argument to `Executor::execute` instead of `__init__` to allow reusing executors with different runners
- When using sequential runners, full runner output (parsed from CSV file) is given as input to the next runner. This way the next runner can access sampled parameters and also previous `run_dir`, `output`, and `success` values
- Added `SumRunner` as an example runner for sequential runs. It is given a list of parameter names in config and on single run it outputs the sum of the values of those parameters.


vanhalaj:
Updated PR description with changed things so far.

#### TODO:

- ~~Update documentation to include MRSAL~~
- ~~Unit tests~~
- ~~Workflow test for nested+sequential run~~
- ~~Seamless sampling in sequential runs?~~
  - I think the current implementation does something like this (didn't test yet):
    - Eg. we have 4 sequential runners: Run is interrupted during the third runner.
    - When continuing, this interrupted sample batch is ignored completely and started over.
  - So I think the current implementation should be enough?
  - I think its good enough t: @JaaskelainenL 

Anything else that is missing?

tsseppa:
Looks good!

src/enchanted_surrogates/supervisor/supervisor.py
DanHarJor, line 141: I haven't tried running the code, but this seems to be set up for the nested implementation for a static dataset. It only goes through each runner, executor combination once.

Nested sampling:
```
        for depth, group in enumerate(self.nested_groups):
            while group.sampler.has_budget:
                samples = group.sampler.get_next_samples()
```

Sequential sampling:
```
        while sampler.has_budget()
            samples = group.sampler.get_next_samples()            
            for depth, group in enumerate(self.nested_groups):
                # pass samples through each runner, allowing the next runner to see the outut of the previous runner    
```

DanHarJor (commented):
I will try running the code. But to me it seems the multi runner sequential sampling is not implimented.

src/enchanted_surrogates/supervisor/supervisor.py
```diff
@@ -173,23 +149,31 @@ def start(self):
...
                     expanded = samples
 
-                # Create run directories named by depth, batch and sample numbers
-                run_dirs = [
-                    os.path.join(real_run_dir, "data", f"d{depth}_b{batch_number}_r{i}")
-                    for i in range(len(expanded))
-                ]
+                # Run for each sequential runner/executor combination
+                df_batch = pd.DataFrame()
+                for i, (executor, runner) in enumerate(
```
JaaskelainenL, line 154: We already do it here? Also this supports both nested and sequential

src/enchanted_surrogates/supervisor/supervisor.py
JaaskelainenL: We already do that.  Line 154 ish

DanHarJor: Ahh I see. The way it is coded is elegant, but a little confusing. 

I tested it using the random sampler and the friedman case and it seems to be working.

Can you please briefly explain how the different parts of run_order should be made and how this changes the self.nested_groups and the groups within them?

JaaskelainenL: It works like in the diagram sent on slack a week or so ago. the run order is sort of like this:
sequential no nesting (c1 run to ex ->  c2 run to ex -> c3 run to ex ):
  run_order:
  - sampler: code1_sampler
    executor:
      - code1_executor
      - code2_executor
      - code3_executor
    runner:
      - code1_runner
      - code2_runner
      - code3_runner
      
     if you add another element to that list in run_order. you will have a nested run after the sequential one. the one after can also be sequential or a normal one. a nested sequential could look something like this:  run_order:
  - sampler: c1_sampler
    executor:
      - c1_executor
      - c2_executor
      - c3_executor
    runner:
      - c1_runner
      - c2_runner
      - c3_runner
  - sampler: c2_sampler
    executor:
      - c4_executor
      - c5_executor
      - c6_executor
    runner:
      - c4_runner
      - c5_runner
      - c6_runner
      
or if you dont want the runs afterwards/before to be sequential it can just be:
run_order:
  - sampler: c1_sampler
    executor:
      - c1_executor
      - c2_executor
      - c3_executor
    runner:
      - c1_runner
      - c2_runner
      - c3_runner
  - sampler: c2_sampler
    executor:
      - c4_executor
    runner:
      - c4_runner

DanHarJor (approved):
I have tested it with the friedman branch. Both the nested and sequential yaml files are working.

Great work.

PR #215, opened 2026-03-19
Feature/enc 185 pylint is fairly slow

JaaskelainenL:
Move pylint to ruff. It is way faster and it is really customizable with what rules to include and exclude. I think the review should mainly focus on if the current ruleset makes sense. I tried to edit it to work close to how pylint does it, however it seems to be a bit more strict


tsseppa:
LGTM

awlishar:
Please update docs/index.md

JaaskelainenL:
Docs updated

awlishar (approved):
Good!

PR #217, opened 2026-03-21
Feat/enc 176 documentation from the rest of plugins

shcera:
**Changes**

- Added docstrings for the remaining plugins from private repositories to the documentation. For the helena, castor, and mishka plugins, all docstrings are added. For the gene plugin, only the docstrings from gene_parser have been added, as other files are missing or incomplete docstrings. Does it make sense to add the entire module?

- Adding docstring documentation for plugins is done through the plugins.yml configuration file. If new plugins are added in the future whose docstring documentation needs to be imported into the main plugin, this can be done simply by updating the configuration file as follows:
```yml
- name: template_plugin
repo: DIGIfusion/enchanted-plugin-template
module: enchanted_plugin_template
src_path: src
```

- Created instructions; see the documentation.md file.


.github/workflows/documentation.yml
```diff
@@ -24,19 +24,38 @@ jobs:
...
+            mkdocs-get-deps==0.2.2 \
+            mkdocs-material-extensions==1.3.1 \
+            mkdocstrings==1.0.3 \
+            mkdocstrings-python==2.0.3 \
+            markdown==3.10.2 \
+            pymdown-extensions==10.21 \
+            jinja2==3.1.6 \
+            pygments==2.19.2 \
+            pyyaml==6.0.3 \
+            h5py==3.16.0
```
JaaskelainenL: LGTM but is there a reason why the versions have to be specified here? do they break on newer ones?

JaaskelainenL:
LGTM

.github/workflows/documentation.yml
shcera: The newest version of mkdocs 2.0, which not realised yet, but will release in future probably break existing plugin system (I mean plugin as mkdocstring for generation documentation from docstrings). Generation from docstrings was the main feature. Specifying only mkdocs versions <2.0 is possible, but how will new plugin versions interact with the old version? To avoid surprises when a new version is released and the site crashes in the future, the versions were specified. 
I will study this issue again and think about whether it is worth specifying all versions.

awlishar (approved):

Issue #201, opened 2026-03-10
bug: enchanted_dataset file is overwritten not appended to


DanHarJor:
In develop the, start() function write_summary() is currently overwriting enchanted_dataset.csv with the most recent batch and then making another redundant enchanted_complete_dataset at the end. 

There should be a single enchanted_dataset.csv containing the data from all the batches, not just a single batch.

Each batch should be **appended** to the csv.

Also it should only be compressed into a parquet file at the very end. When the entire enchanted_dataset.csv has had all the batches appended, then it should be compressed into a parquet file, if the user specified to with the config file. This is because appending to a parquet file is not easy and expensive. Compressing to parquet is expensive and should only be done once. (the benefit is smaller sizes and faster data loading)

awlishar:
Parquet file compresssion optional

PR #218, opened 2026-03-23
Feature/enc 168 bug enchanted dataset file is overwritten not appended to

tsseppa:
Now in the start() function loop the `enchanted_dataset.csv` is appended, not overwritten. Parquet is created in the very end. 

I am not sure if with the [issue](https://github.com/DIGIfusion/enchanted-surrogates/issues/201) was meant that there should not be last_completed_dataset in the loop at all, so I left it there as it is quite crucial for there to have seamless sampling for nested runs if I understood it correctly.


JaaskelainenL:
LGTM

awlishar (approved):

PR #220, opened 2026-03-26
fix: renamed plugin token to match current

tsseppa:
Renamed plugin to match the current one, to enable build of documentation to succeed.


shcera (approved):
Approved

awlishar (approved):

PR #222, opened 2026-03-26
CI: cached workflows

JaaskelainenL:
Reuse the same venv for workflows. This should potentially speed up the workflows by ~10-20s.


vanhalaj:
Looks good! Automated testing total time went from ~85 seconds to ~60 seconds

awlishar (approved):

PR #228, opened 2026-03-31
Feature/enc 205 pt change default logging to info

JaaskelainenL:
Another peer review issue. Debug has too much info especially with the dask stuff. Info is a better default. Developers can change it to debug when needed.


tsseppa:
LGTM

awlishar (approved):

PR #229, opened 2026-04-01
bug: removing files with save_files not working correctly

dafughes:
Currently, when setting config option `save_files` to "none" (or "custom"):
- Supervisor fails to remove the log files (since they are locked by the logging process)
- Creating the hdf5 file fails, since `data` directory is already removed.

This PR fixes these by:
- Ignoring `logs` dir when deleting files (file logging could be made a config option)
- Updating the hdf5 file after each batch run

### DIFF
--- a/docs/supervisor.md
+++ b/docs/supervisor.md
@@ -68,32 +68,34 @@ See [example_nested.yaml](../configs/example_nested.yaml).
 
 ### Multi-runner sequential execution
 
-Alongside nested executing, the supervisor also supports sequential sampling.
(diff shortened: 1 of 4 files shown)


vanhalaj:
Looks good!

Similar to LOG_DIR here, I think we should make other currently hard-coded values (summary file name etc) in `Supervisor` constants as well. I will create separate issue for this

JaaskelainenL:
LGTM

awlishar (approved):

PR #233, opened 2026-04-10
Fix: BO sampler duplicate observations

awlishar:
Issue: Running the BO example case with the example config resulted in the samples being added twice to y_obs and x_obs. Which then resulted in only half the number of initial samples and bad convergence.

Solution: The base_run_dir in the BO sampler config was defined as the same directory as the base_run_dir in the supervisor, will result in the samples being added twice to the dataset. Removed base_run_dir from sampler config.

Other changes: The BO sampler maximizes the objective (distance). So I changed the parser to output -(y_target - f(x)) instead of multiplying the objective by -1 in the sampler.


DanHarJor:
I think it makes sense for the sampler to handle this part -(y_target - f(x)). This way it is implemented once rather than every runner needing to implement it if they want to do BO.

awlishar:
> I think it makes sense for the sampler to handle this part -(y_target - f(x)). This way it is implemented once rather than every runner needing to implement it if they want to do BO.

I made a new issue for that, I can implement it as a separate PR.

DanHarJor (approved):
Looks good to me

PR #240, opened 2026-04-13
Daniel Stacked PR [1] bug example_runner for nested workflows

DanHarJor:
I noticed an error in example_runner that exists when doing nested runs. The c1 and c2 would be written to the wrong parameters as they are written to the first ones in params. Which in nested runs is 'output' of the previous run. So it now selects the parameters with the largest number in the parameter key. Eg c1:1, c2:1, c3:3, c4:4, it would take c3 and c4.


DanHarJor:
I is better the run_simulation_task is not altered and the functionality of having the failed runs logged as an option in the supervisor.

awlishar:
What is the difference between files all_progress and current_progress?

DanHarJor:
In the future I would like just one progress report file, this is a little more difficult. I made an issue about it. 
Here is an example of current progress:

=== PROGRESS REPORT =====================================

Time:            2026-04-15 10:43:39.669769
Depth:           0
Batch:           0

Completed:       6/9
Successes:       0
Failures:        6
Success Rate:      0.0%

[##########################--------------]  66.7%

==========================================================

This is only for the current depth and batch. I made an issue to make a file for the entire pipeline progress. This is too big a task for right now.

This file is overwritten so when I got the next batch or depth, I cannot see the success rate of the previous ones, so I append them into all progress:


        
=== PROGRESS REPORT =====================================

Time:            2026-04-15 10:59:48.605389
Depth:           0
Batch:           0

Completed:       9/9
Successes:       2
Failures:        7
Success Rate:     22.2%

[########################################] 100.0%

==========================================================


        
=== PROGRESS REPORT =====================================

Time:            2026-04-15 11:00:02.461331
Depth:           1
Batch:           0

Completed:       36/36
Successes:       31
Failures:        5
Success Rate:     86.1%

[########################################] 100.0%

==========================================================

awlishar:
Okey please update the documentation: https://digifusion.github.io/enchanted-surrogates/#output

awlishar (approved):
:+1:

PR #242, opened 2026-04-13
Daniel Stacked PR [2] feature/monitor_runs_log_failures

DanHarJor:
When there are a lot of dask workers, it can be tedious to check all their separate log files for the reasons why the runs are failing.

This is why I added a monitor to check for failed runs and log the failures in the main log file. This allows user to easily see why their runs are failing as they fail.


DanHarJor:
I will make it configurable and add a quick and basic progress monitor file.

PR #244, opened 2026-04-14
Fix active sampling candidates

awlishar:
The number of candidates was set to `batch_size`. Meaning that the ML model only can chose from that number of candidates. This results in the same candidates being chosen.

The number of candidates is now a configurable input. It should be significantly larger than the budget. If budget>=n_candidates then we just get random sampling, no active learning. Once a candidate is chosen, it is removed from the pool of available candidates.


DanHarJor (approved):
Looks great, thanks for looking over it.

PR #256, opened 2026-04-17
Feature/ensure below budget

DanHarJor:
this ensures that enchanted surrogates stays below budget if the next samples is very large then it can go very over budget without this fix. It also allows samplers control by returning none when it wants to stop sampling for any reason


DanHarJor:
ruff check .
All checks passed!

awlishar (approved):

PR #261, opened 2026-04-21
Feature/single beautiful progress file

DanHarJor:
I renamed the depth variables so it was clear exactly what they are. I changed the run dir naming so it is clear what each part is. I made the progress report file much cleaner and now there is only one. It will show the entire progress for sequential runs. For nested runs it will only show the progress for the current nested group


src/run.py
```diff
@@ -18,49 +19,6 @@
 """
 
 
-def load_configuration(config_path: str) -> argparse.Namespace:
```
DanHarJor, line 21: I needed to move these out of the run file so that they can be imported from the enchanted-surrogates submodule for the testing. This allows the relative import to be removed from the tests which was breaking the tests when I was trying to run them locally.

DanHarJor:
ruff check .
All checks passed!

PR #262, opened 2026-04-21
fix-only_delete_in_datadir

DanHarJor:
My samplers might put files in the base_run_dir, I do not want them to be deleted just because I forgot to put them in the config. 

The main purpose of this delete functionality was to avoid too many files error. The data dir is the only one at real risk of this. 

Now delete unwanted files will only affect the data_dir.


DanHarJor:
ruff check .
All checks passed!

awlishar (approved):

PR #266, opened 2026-04-28
bug/supervisor_registering_too_much

DanHarJor:
the supervisor was registering the entire dataset every batch. It should only register the new data from the new batch.


DanHarJor:
ruff check .
All checks passed!

awlishar (approved):

PR #268, opened 2026-05-15
Fix/only success allowed

DanHarJor:


DanHarJor:
ruff check .
All checks passed!

awlishar (approved):

PR #282, opened 2026-09-08
Route SLURM worker stdout/stderr into the run's log directory

DanHarJor:
dask_jobqueue defaults slurm-<jobid>.out files to the submission cwd, cluttering the launch directory. Point them at the same log_directory used for per-worker application logs instead.


aejarvin (approved):
Looks good.

PR #283, opened 2026-09-08
Feature/loglinear grid sampler

DanHarJor:
added the ability to sample on a log scale along some or all of the dimensions if desired.


aejarvin (approved):
This feature should not conflict with any of the existing code. Looks good to me.

PR #284, opened 2026-09-08
Urgent/fix precise imports monkeypatch string path

DanHarJor:
The test were failing for all my PR's even for very minor changes. The test would work on LUMI but fail on GitHub. 

I asked claude for a fix and this was the suggestion.


aejarvin (approved):
Pretty simple change. Let's merge and then fix if needed.

PR #286, opened 2026-09-08
Feature/post processing

DanHarJor:
allows the user to add a script that is automatically ran on the scheduler node within the base run dir after all the simulations are finished. This way thay can add some plotting or surrogate model training or anything they need to do with the dataset automatically.


aejarvin (approved):
Looks like small changes that do not conflict.

PR #288, opened 2026-09-11
Aaro test/data packers clean when finished

DanHarJor:
a merge of the last two PR's so Aaro can test them both in one go.


aejarvin (approved):
Test ran through without issues.

